PR #39, opened 2021-08-05
Adds `RestoreAsOf` and `Previous` fields to the restic data-mover

osilkin98:

This PR adds `Previous` and `RestoreAsOf` fields to the `ReplicationDestinationResticSpec`. 

`RestoreAsOf` - Serves as an anchor when selecting a Restic snapshot to restore from, and expects a string provided as an `RFC 3339` timestamp. When `RestoreAsOf` is provided, the restic data mover will use it as the starting point when looking for timestamps. When left blank, it will default to `Time.Now()` during synchronization. 

`Previous` - specifies the offset when selecting a restic snapshot to restore from relative to the provided anchor. For example, if `RestoreAsOf` is set to the current time, the `Previous` oldest snapshot will be selected. If `RestoreAsOf` is set to sometime a couple of days ago, the `Previous` oldest snapshot at that time will be selected.

**Related issues:**


cooktheryan:
We should be able to use kuttl to monitor the status of a volume snapshot. Maybe that will help us

osilkin98:
@JohnStrunk I had to disable parallelism since the restic e2e tests would write snapshots to the same minio bucket which created a race condition where tests would restore from each other's snapshot, subsequently failing the tests.

JohnStrunk:
> @JohnStrunk I had to disable parallelism since the restic e2e tests would write snapshots to the same minio bucket which created a race condition where tests would restore from each other's snapshot, subsequently failing the tests.

What about using a different bucket for the different tests? The bucket name is defined in the restic-repo secret: 
https://github.com/backube/volsync/pull/39/files#diff-d7538463a216881fca23cc83266922f36577991f55793365736cf8987e92c6d1R16
If you name a bucket that doesn't exist, it'll get created.

cooktheryan:
awful lot of green check marks there :+1:

mover-restic/entry.sh
```diff
@@ -72,12 +76,156 @@ function do_prune {
...
+            fi
+            ((offset++))
+        done
+    fi
+
+    # offset the epoch selection if SELECT_PREVIOUS is defined
+    local -i select_offset=${SELECT_PREVIOUS-0}
+    ((offset+=select_offset))
+
+    # when timestamp isnt provided, just retrieve the latest snapshot
```
JohnStrunk, line 194: This comment really confused me. When timestamp isn't provided, offset is 0, so we'll just take the most recent.
This block is really just converting the offset to the snapshot id, right?

test-kuttl/e2e/restic-with-restoreasof/22-recreate-restore.yaml
```diff
@@ -0,0 +1,7 @@
+# issues a command to recreate the restic snapshot
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - timeout: 2400
```
JohnStrunk, line 6: Why would this step take a long time?

test-kuttl/e2e/restic-with-restoreasof/23-verify-data.yaml
```diff
@@ -0,0 +1,27 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: verify
+spec:
+  backoffLimit: 99
```
JohnStrunk, line 7: Is there a need to retry?

test-kuttl/e2e/restic-with-restoreasof/23-assert.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+timeout: 2400
```
JohnStrunk, line 4: Could this be something less... like 600 or 300?

JohnStrunk (commented):
quite the bash :ninja: 

Just a few small items, nothing major.

controllers/mover/restic/mover.go
```diff
@@ -264,7 +268,7 @@ func (m *Mover) ensureJob(ctx context.Context, cachePVC *corev1.PersistentVolume
 		}
 		utils.MarkForCleanup(m.owner, job)
 		job.Spec.Template.ObjectMeta.Name = job.Name
-		backoffLimit := int32(8)
+		backoffLimit := int32(999999999)
```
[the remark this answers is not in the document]
osilkin98: yep

mover-restic/entry.sh
osilkin98: This comment was forgotten after this function was drastically changed. A Classic example of code debt.

test-kuttl/e2e/restic-with-restoreasof/22-recreate-restore.yaml
osilkin98: In the conversation the other day it was mentioned that we needed to bump up the timeouts, at the time I was under the impression that it was the cause of my e2e test failures.

JohnStrunk:
Thanks for working on this.
/approve
/lgtm

PR #42, opened 2021-08-05
Add ability to specify full container image in Chart

### LINKED ISSUE #5: Allow helm install to use sha hashes of container images
Currently, the helm chart allows specifying the container image and a tag name. It should also be possible to specify a sha hash instead of a named tag.

We need to use the hashes in openshift ci builds, and currently it has to be hacked around.

**Additional context**
cc: @sallyom

JohnStrunk:
This adds an additional field to the Chart values file that allows specifying the full image field unstead of individually specifying the image and tag. The purpose is to permit specifying the image+sha hash.

Note:
`x.image` overrides both `x.repository` and `x.tag`

**Related issues:**
Fixes #5

### DIFF
--- a/helm/volsync/templates/_helpers.tpl
+++ b/helm/volsync/templates/_helpers.tpl
@@ -60,3 +60,19 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Determine the container image to use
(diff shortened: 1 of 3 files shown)


JohnStrunk:
@sallyom PTAL

sallyom:
/lgtm

When this merges I'll update the [e2e PR](https://github.com/openshift/release/pull/20036) to be sure all is working as expected.

PR #47, opened 2021-08-12
Race in set-replication

### LINKED ISSUE #14: Race in `set-replication`
When set-replication performs the manual sync, it waits for the sync to finish on the source (trigger status == spec), then it immediately attempts to retrieve the `latestImage` from the ReplicationDestination. There is no guarantee that `latestImage` has been updated yet, even though we know the sync has completed from the PoV of the source.

Result in the following error being triggered:
https://github.com/backube/scribe/blob/a9c0d0c9d4084c3aeaa61f771c19b8d736acd8a5/pkg/cmd/set_replication.go#L197

This may be a bit tricky to solve since the mere presence of a `latestImage` is insufficient (it may be from a previous sync). We somehow need to poll and ensure what we find is generated after the completion of the sync. Perhaps:
1. Check destination and remember what we find
2. Do the manual sync on the source
3. Poll the destination until it's different than what we saw in step 1 (or timeout)

v-harihar:
Issue:
When set-replication performs the manual sync, it waits for the sync to finish
on the source (trigger status == spec), then it immediately attempts to retrieve
the latest images from the ReplicationDestination. There is no guarantee that
the latest image has been updated yet, even though we know the sync has been 
completed from the PoV of the source.

Solution:
1. Check destination and remember what we find
2. Do the manual sync on the source
3. Poll the destination until it's different than what we saw in step 1 (or timeout)

**Related issues:**
Fixes #14

### DIFF
--- a/pkg/cmd/set_replication.go
+++ b/pkg/cmd/set_replication.go
@@ -150,11 +150,28 @@ func (o *FinalizeOptions) SetReplication() error {
 	if err := o.RepOpts.Source.Client.Get(ctx, sourceNSName, repSource); err != nil {
 		return err
 	}
+
+	// Keep the oldImage saved to for the reference
+	var oldImage *corev1.TypedLocalObjectReference
+	repDest := &volsyncv1alpha1.ReplicationDestination{}
+	destNSName := types.NamespacedName{
+		Namespace: o.RepOpts.Dest.Namespace,
+		Name:      o.destName,
+	}
+
+	if err := o.RepOpts.Dest.Client.Get(ctx, destNSName, repDest); err != nil {
+		return err
+	}
+	if repDest.Status != nil {
+		oldImage = repDest.Status.LatestImage
+	}
+
 	klog.Infof("Triggering final data sync")
 	repSource.Spec.Trigger = &volsyncv1alpha1.ReplicationSourceTriggerSpec{
 		Schedule: repSource.Spec.Trigger.Schedule,
 		Manual:   lastManualSync,
 	}
+
 	if err := o.RepOpts.Source.Client.Update(ctx, repSource); err != nil {
 		return fmt.Errorf("unable to set manual trigger for last sync")
 	}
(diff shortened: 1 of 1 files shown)


pkg/cmd/set_replication.go
```diff
@@ -155,6 +155,18 @@ func (o *FinalizeOptions) SetReplication() error {
...
+	repDest := &volsyncv1alpha1.ReplicationDestination{}
+	destNSName := types.NamespacedName{
+		Namespace: o.RepOpts.Dest.Namespace,
+		Name:      o.destName,
+	}
+	if err := o.RepOpts.Dest.Client.Get(ctx, destNSName, repDest); err != nil {
+		return err
+	}
+
+	oldImage := repDest.Status.LatestImage
```
JohnStrunk, line 168: You'll need to check `repDest.Status` for nil before dereferencing
Probably something like:
```go
var oldImage *corev1.TypedLocalObjectReference
if repDest.Status != nil {
  oldImage = repDest.Status.LatestImage
}
```

pkg/cmd/set_replication.go
```diff
@@ -175,19 +187,33 @@ func (o *FinalizeOptions) SetReplication() error {
...
-	if err := o.RepOpts.Dest.Client.Get(ctx, destNSName, repDest); err != nil {
+
+	if err := wait.PollImmediate(5*time.Second, o.timeout, func() (bool, error) {
+		if err := o.RepOpts.Dest.Client.Get(ctx, destNSName, repDest); err != nil {
+			return false, err
+		}
+		latestImage = repDest.Status.LatestImage
+		if oldImage.Name == latestImage.Name {
+			return false, nil
+		}
```
JohnStrunk, line 209: - Need to check that `Status` isn't nil
- Need to check that `oldImage` and `latestImage` aren't nil

JohnStrunk (changes requested):
Thanks for taking a look at this!

JohnStrunk:
Thanks, Vinayak

/lgtm

JohnStrunk (approved):

Issue #49, opened 2021-08-12
cli should be usable w/o a config file


JohnStrunk:
**Describe the feature you'd like to have.**
It should be possible to use the cli w/o having a config file.

Currently:
```console
$ kubectl volsync help
error: Config File "config.yaml" Not Found in "[/home/jstrunk/.volsyncconfig]"
```

**What is the value to the end user? (why is it a priority?)**

**How will we know we have a good solution? (acceptance criteria)**
- A missing config file should be treated as empty
- ~It should be possible to provide an alternate location for the config file~ This exists `--config <filename`
- It should be possible to use the cli in full w/o a config file (everything via cli parameters)

**Additional context**
Related to #22, #10

sallyom:
@JohnStrunk I only noticed this the other day, **this is a bug** not intended behavior,  thank you!

PR #51, opened 2021-08-17
cli: set-replication doesn't obey source-name

### LINKED ISSUE #10: cli docs show incorrect option names
**Describe the bug**
The option names described in the docs don't match what the cli expects.
docs: https://scribe-replication.readthedocs.io/en/latest/usage/rsync/plugin_opts.html

At least:
- dest-storage-class-name -> dest-storage-class
- dest-volume-snapshot-class-name -> dest-volume-snapshot-class

### LINKED ISSUE #50: cli: set-replication doesn't obey `source-name`
**Describe the bug**
`start-replication` allows creating a custom named ReplicationSource via the `source-name` parameter. `set-replication` should also use this parameter since it's commonly used as part of a start, set sequence.

**Steps to reproduce**
Config file:
```yaml
source-kube-context: kind-kind
source-name: wiki-src
source-namespace: wiki
source-copy-method: Clone
source-pvc: dokuwiki

dest-kube-context: admin
dest-name: wiki-dest
dest-namespace: wiki-dest
dest-service-type: LoadBalancer
dest-capacity: 2Gi
dest-access-mode: ReadWriteOnce
dest-copy-method: Snapshot
```

```console
$ kubectl volsync start-replication
...
$ kubectl volsync set-replication
I0813 10:27:22.687209  523925 set_replication.go:144] Fetching ReplicationSource wiki-source in namespace wiki
Error from server (NotFound): replicationsources.volsync.backube "wiki-source" not found

$ kubectl --context kind-kind -n wiki get replicationsource
NAME       SOURCE     LAST SYNC              DURATION        NEXT SYNC
wiki-src   dokuwiki   2021-08-13T14:30:19Z   19.081158175s   2021-08-13T14:35:00Z
```

**Expected behavior**

**Actual results**
`set-replication` tried to use "wiki-source" instead of "wiki-src"

**Additional context**

v-harihar:
start-replication allows creating a custom named ReplicationSource
via the source-name parameter. set-replication should also use this
parameter since it's commonly used as part of a start, set sequence.

Due to different source and destination names used for command line
parameters in case of start-replication and set-replication, the
set-replication command tries to find the replication destination
and replication source in default namespaces and fails.

corrected command-line parameters as mentioned in
https://scribe-replication.readthedocs.io/en/latest/usage/rsync/plugin_opts.html

**Related issues:**
Fixes #50
Fixes #10

### DIFF
--- a/pkg/cmd/set_replication.go
+++ b/pkg/cmd/set_replication.go
@@ -243,16 +243,17 @@ func (o *FinalizeOptions) SetReplication() error {
 		destOpts := ResourceOptions{
 			PVC: o.destPVC,
 		}
+		sourceOpts := ResourceOptions{
+			PVC:          srcPVC.Name,
+			StorageClass: o.destStorageClass,
+		}
 		repOpts := &SetupReplicationOptions{
 			RepOpts: o.RepOpts,
 			Dest: DestinationOptions{
 				destOpts,
 			},
 			Source: SourceOptions{
-				ResourceOptions: ResourceOptions{
-					PVC:          srcPVC.Name,
-					StorageClass: o.destStorageClass,
-				},
+				sourceOpts,
 			},
 		}
 		if len(o.destCapacity) == 0 {
--- a/pkg/cmd/set_replication.go
+++ b/pkg/cmd/set_replication.go
@@ -89,12 +89,13 @@ func NewCmdVolSyncSetReplication(streams genericclioptions.IOStreams) *cobra.Com
 
 func (o *FinalizeOptions) bindFlags(cmd *cobra.Command, v *viper.Viper) error {
 	flags := cmd.Flags()
-	flags.StringVar(&o.sourceName, "source-replication-name", o.sourceName, "name of ReplicationSource (default '<source-ns>-source')")
-	flags.StringVar(&o.destName, "dest-replication-name", o.destName, "name of ReplicationDestination (default '<dest-ns>-destination') ")
+
+	flags.StringVar(&o.sourceName, "source-name", o.sourceName, "name of ReplicationSource (default '<source-ns>-source')")
+	flags.StringVar(&o.destName, "dest-name", o.destName, "name of ReplicationDestination (default '<dest-ns>-destination') ")
 	flags.StringVar(&o.destPVC, "dest-pvc", o.destPVC, "name of not-yet-existing destination PVC. "+
 		"Default is sourcePVC name, or if PVC with sourcePVC name exists in destination namespace, then 'sourcePVC-<date-tag>'")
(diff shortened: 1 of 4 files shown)


pkg/cmd/set_replication.go
```diff
@@ -89,7 +89,14 @@ func NewCmdVolSyncSetReplication(streams genericclioptions.IOStreams) *cobra.Com
 
 func (o *FinalizeOptions) bindFlags(cmd *cobra.Command, v *viper.Viper) error {
 	flags := cmd.Flags()
+
+	if o.sourceName == "" {
+		o.sourceName, _ = v.Get("source-replication-name").(string)
+	}
 	flags.StringVar(&o.sourceName, "source-replication-name", o.sourceName, "name of ReplicationSource (default '<source-ns>-source')")
```
JohnStrunk, line 96: It looks like you're trying to update the default value for the flags. I think a better approach would be to ensure the flag names are standardized across the different operations.

For example, the current issue of `start-replication` vs `set-replication` looks like it comes down to the flags being named differently (and not documented).
`start-replication` seems to take "source-name" (L126 of create_replication.go) whereas `set-replication` takes "source-replication-name".

My suggestion would be to take a look at the flag names and adjust them so that they are consistent across the different sub-commands. It would also be beneficial to ensure consistent naming of the struct fields, too. (i.e. `o.SourceName` vs `o.Name`)

A good review and cleanup could probably take care of #10 and #49 as well.

v-harihar: Makes sense. I too came across these differences but was not sure whether to take of those changes through this PR. 
I will do the needful.

v-harihar:
@JohnStrunk #49 is not handled in this commit.

pkg/cmd/set_replication.go
```diff
@@ -239,14 +240,20 @@ func (o *FinalizeOptions) SetReplication() error {
...
 		}
 		repOpts := &SetupReplicationOptions{
-			RepOpts:      o.RepOpts,
-			DestOpts:     destOpts,
-			SourcePVC:    srcPVC.Name,
-			StorageClass: o.destStorageClass,
+			RepOpts: o.RepOpts,
+			Dest: DestinationOptions{
+				destOpts,
+			},
```
sallyom, line 250: nit: Inline the destOpts (L243,244) like below with SourceOptions, for consistency, (or define sourceOpts above, like destOpts) but I'm fine with it as/is

sallyom (commented):
@v-harihar thank you so very much for the much needed cleanup and fixes!
This lgtm, I have the one nit, but other than that let's merge this and keep improving on the CLI  :grin:

pkg/cmd/set_replication.go
v-harihar: Sure

Issue #48, opened 2021-08-12
When allocating a new volume based on another, use the correct capacity


JohnStrunk:
When generating a clone of an existing PVC, we currently use the requested size:
https://github.com/backube/volsync/blob/216ac45a8ac7ae851d7a1713fec16a0a8a7cd459/controllers/volumehandler/volumehandler.go#L243
https://github.com/backube/volsync/blob/216ac45a8ac7ae851d7a1713fec16a0a8a7cd459/controllers/volumehandler/volumehandler.go#L346
https://github.com/backube/volsync/blob/216ac45a8ac7ae851d7a1713fec16a0a8a7cd459/controllers/volumehandler.go#L342
https://github.com/backube/volsync/blob/216ac45a8ac7ae851d7a1713fec16a0a8a7cd459/controllers/volumehandler.go#L430
We should be using the actual size (`.status.capacity`). These two could differ in the case of volume resize.

PR #64, opened 2021-08-26
When allocating a new volume based on another, use the correct capacity

v-harihar:
When generating a clone of an existing PVC, we currently use the requested
size. So We should be using the actual size (.status.capacity). These two
could differ in the case of volume resize.

Fixes: #48


tesshuflower:
I think this PR would have fixed the issue, but got out of date.  We ended up fixing in https://github.com/backube/volsync/pull/251 and had slightly more checks (i.e. don't rely on capacity, still fall back to requested size etc).  I think we can probably close this one now.

PR #65, opened 2021-08-26
Adds moving image tags for release & ocm branches

JohnStrunk:
This adds moving tags to the container images in Quay that will track
the HEAD of the release and ocm branches.

**Related issues:**
open-cluster-management/backlog#14604


JohnStrunk:
/cherrypick ocm-2.4

Looks like the workflow files need to be in the target branch for them to run

Issue #73, opened 2021-09-07
Build rclone and restic binaries


JohnStrunk:
**Describe the feature you'd like to have.**
Today, we download the released binaries for rclone and restic to include in the mover containers. This presents an auditing problem for the downstream build & release process.
We should clone and build the sources ourselves as a part of the Docker build of the mover containers.

**What is the value to the end user? (why is it a priority?)**
- This ensures reproducible builds
- Makes it easier to manage upstream/downstream releases and keeps them better in sync
- Makes it easier to build for multiple architectures

**How will we know we have a good solution? (acceptance criteria)**
- The upstream source is cloned & built on each container build
- The build is locked to a specific release by tag and git hash
- Version information is visible in the Dockerfile such that a simple diff of the file can determine if/when the version changed (i.e., the version shouldn't be a `buildarg`)

**Additional context**
Related to open-cluster-management/backlog#14604

JohnStrunk:
This is Fixed by #74 & #75

PR #74, opened 2021-09-07
Build restic and rclone from source

JohnStrunk:
This PR builds the rclone and restic binaries from source instead of relying on the built binary releases.

**Related issues:**
Related to #73

### DIFF
--- a/mover-rclone/Dockerfile
+++ b/mover-rclone/Dockerfile
@@ -1,3 +1,22 @@
+# Build rclone
+FROM golang:1.16 as rclone-builder
+
+WORKDIR /workspace
+
+ARG RCLONE_VERSION=v1.55.1
+# hash: git rev-list -n 1 ${RCLONE_VERSION}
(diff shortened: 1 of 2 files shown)


JohnStrunk:
@arewm Could you take a look and make sure this fits what you were expecting

arewm (approved):
LGTM! Thanks!

When porting this downstream, the git-ness of the repo is cleaned out. Do you know if any additional parameters are needed to pass to the complilation to rectify this (i.e. to output version metadata, etc. to logging)? 

No worries if you don't know _now_ as we can iterate on it in the future.

JohnStrunk:
@arewm We embed the version for the operator based on this make variable. It can be overridden:
https://github.com/backube/volsync/blob/744cf2e1f3cfa6f1f92ca96333138dddf8a46a46/Makefile#L6

There's also a VERSION variable in the movers that's set the same way.

PR #79, opened 2021-09-21
Clean up e2e... again

JohnStrunk:
Hopefully this improves e2e pass rate on OCP/AWS.

- It increases the default step timeout to 300s and manually increases to 600s when snapshots will be accessed. Even for the small volumes we use in e2e, it can take several minutes (typically ~2.5m) for a snapshot to become ready. This long delay has the effect of causing the provisioner to backoff, causing another 1m or so delay once `ready` is achieved.
- This also fixes a bug in restoreasof where the "source" pod wasn't being deleted, causing a multi-attach error when the verify pod tried to start.
- It adds additional data collection when a job fails to help diagnosing
- It pulls the separate *.yaml and *.sh test scripts into the yaml via embedded scripts, so the test step is all in one place.

**Related issues:**


test-kuttl/e2e/restic-with-manual-trigger/00-create-restic-secret.yaml
```diff
@@ -2,5 +2,23 @@
 apiVersion: kuttl.dev/v1beta1
 kind: TestStep
 commands:
-  - timeout: 90
-    command: ./00-create-restic-secret.sh
+  - timeout: 60
+    script: |
```
osilkin98, line 6: Nice usage of `script`, removing the extra `.sh` scripts is a great step.

osilkin98 (approved):
This looks good, it's nice that a lot of the default timeouts have been removed, that'll make it easier to increase timeouts if we need to in the future.

PR #80, opened 2021-09-22
Add VersionInfo to Builder interface

### LINKED ISSUE #78: Print mover container images at startup
When the operator starts, it should print the container image names/tags that will be used for the movers. Currently, this works for rsync & rclone, but the image path for restic (which uses the new mover interface) isn't printed.

Enhance the Mover interface such that the container image name gets printed to the log when a mover is added to the catalog.

It should be easy to identify exactly which container images are used to help troubleshooting.

- [x] The full container image name + tag are printed for all movers
- [x] There exists a standard way of doing this using the Mover interface

**Additional context**

JohnStrunk:
The VersionInfo() call allows movers to print the container image that they will be using. This is useful information in the debug logs.

**Related issues:**
Fixes #78


osilkin98 (approved):
/lgtm

PR #90, opened 2021-09-30
CLI: Migration create

v-harihar:
This sub-command should prepare a destination to receive incoming transfers from an external source.

- [ ] Saves relationship info
- [ ] Creates the Namespace if it doesn't exist
- [ ] Creates a PVC in the Namespace if it doesn't exist
- [ ] Using the specified capacity, accessModes, StorageClass, and name
- [ ] Creates an associated ReplicationDestination

**Related issues:**

### DIFF
--- a/kubectl-volsync/cmd/migration_create.go
+++ b/kubectl-volsync/cmd/migration_create.go
@@ -17,13 +17,57 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 package cmd
 
 import (
+	"context"
 	"fmt"
 
+	volsyncv1alpha1 "github.com/backube/volsync/api/v1alpha1"
 	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	kerrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/klog/v2"
 	"k8s.io/kubectl/pkg/util/i18n"
 	"k8s.io/kubectl/pkg/util/templates"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+type DestinationInfo struct {
+	Name         string
+	Namespace    string
+	Opts         VolSyncOptions
+	Schedule     string
+	CopyMethod   string
+	Capacity     string
+	StorageClass string
+	AccessMode   string
+	PVC          string
+	ServiceType  string
+}
+
+type VolSyncOptions struct {
+	KubeContext     string
+	KubeClusterName string
+	Namespace       string
+	Client          client.Client
+	CopyMethod      volsyncv1alpha1.CopyMethodType
+	Capacity        resource.Quantity
+	StorageClass    *string
+	SSHUser         *string
+	ServiceType     corev1.ServiceType
+}
+
+type SourceInfo struct {
+}
+
+type MigrationParams struct {
+	Dest   DestinationInfo
+	Source SourceInfo
+
+	genericclioptions.IOStreams
+}
 
 // migrationCreateCmd represents the create command
 var migrationCreateCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


kubectl-volsync/cmd/migration_create.go
```diff
@@ -74,24 +123,163 @@ func doMigrationCreate(cmd *cobra.Command, args []string) error {
...
+			APIVersion: "v1",
+			Kind:       "Namespace",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: migParams.Dest.Namespace,
+		},
+	}
+
+	klog.V(2).Infof("Creating NameSpace %s ", migParams.Dest.Namespace)
+	if err := migParams.Dest.Opts.Client.Create(ctx, ns); err != nil {
```
JohnStrunk, line 280: Filter out the error case where the NS exists here (`kerrs.IsAlreadyExists(err)`) before returning.

kubectl-volsync/cmd/migration_create.go
```diff
@@ -74,24 +123,163 @@ func doMigrationCreate(cmd *cobra.Command, args []string) error {
...
+		},
+		ServiceType: &migParams.Dest.Opts.ServiceType,
+	}
+
+	if destPVCExists {
+		rsyncSpec.DestinationPVC = &migParams.Dest.PVC
+	}
+
+	rd := &volsyncv1alpha1.ReplicationDestination{
+		TypeMeta: metav1.TypeMeta{
```
JohnStrunk, line 180: Don't set TypeMeta

kubectl-volsync/cmd/migration_create.go
```diff
@@ -74,24 +123,163 @@ func doMigrationCreate(cmd *cobra.Command, args []string) error {
...
+
+	err = CreateDestination(ctx, migParams, destPVCExists)
+	if err != nil {
+		return fmt.Errorf("%w", err)
+	}
+
+	return nil
+}
+
+func CreateDestination(ctx context.Context, migParams *MigrationParams, destPVCExists bool) error {
```
JohnStrunk, line 163: If the destination PVC doesn't exist, we need to create it so that the name matches what the user provided on the cli. By the time we get here, we should have already done that. No need for `destPVCExists`

kubectl-volsync/cmd/migration_create.go
```diff
@@ -74,24 +123,163 @@ func doMigrationCreate(cmd *cobra.Command, args []string) error {
...
+	qCap, _ := resource.ParseQuantity(migParams.Dest.Capacity)
+	accessModes := []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce}
+	rsyncSpec := &volsyncv1alpha1.ReplicationDestinationRsyncSpec{
+		ReplicationDestinationVolumeOptions: volsyncv1alpha1.ReplicationDestinationVolumeOptions{
+			CopyMethod:  volsyncv1alpha1.CopyMethodNone,
+			Capacity:    &qCap,
+			AccessModes: accessModes,
```
JohnStrunk, line 170: Since we are providing `destinationPVC`, we shouldn't provide these here. They should be used to create the PVC (if necessary)

kubectl-volsync/cmd/migration_create.go
```diff
@@ -74,24 +123,163 @@ func doMigrationCreate(cmd *cobra.Command, args []string) error {
...
+		rsyncSpec.DestinationPVC = &migParams.Dest.PVC
+	}
+
+	rd := &volsyncv1alpha1.ReplicationDestination{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "volsync.backube/v1alpha1",
+			Kind:       "ReplicationDestination",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "migration-destination",
```
JohnStrunk, line 185: Use the PVC name. that way we can have >1 destination in a given Namespace at a time.

kubectl-volsync/cmd/migration_create.go
```diff
@@ -74,24 +123,163 @@ func doMigrationCreate(cmd *cobra.Command, args []string) error {
...
+	}
+	relation.Set("serviceType", serviceType)
 	if err = relation.Save(); err != nil {
-		return fmt.Errorf("unable to save relationship configuration: %w", err)
+		return nil, fmt.Errorf("unable to save relationship configuration: %w", err)
 	}
 
 	// TODO: Set up objects on the cluster
-	_, _ = newClient("")
+	clientObject, _ := newClient("")
```
JohnStrunk, line 237: Pass the cluster context name to `newClient` (from the PVC name's XClusterName object). That way the client will point to the correct kube cluster and not just the user's default one.

JohnStrunk (commented):
I wasn't sure whether you were ready for feedback. This is just some quick thoughts.

kubectl-volsync/cmd/migration_create.go
```diff
@@ -43,11 +87,14 @@ var migrationCreateCmd = &cobra.Command{
 func init() {
 	migrationCmd.AddCommand(migrationCreateCmd)
 
+	migrationCreateCmd.Flags().String("namespace", "", "namespace to create or use: [context/]namespace/name")
```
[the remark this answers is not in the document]
v-harihar: Apologies for replying late. Your feedback is much needed. I will check and update the patch accordingly.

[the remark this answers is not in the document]
v-harihar: If the pvc does not preexist then a namespace name is required to create pvc. In this case how to fetch the name space name ?

kubectl-volsync/cmd/migration_create.go
v-harihar: sure

kubectl-volsync/cmd/migration_create.go
v-harihar: sure

PR #91, opened 2021-09-30
Ensure golang modules are downloaded, not vendored

JohnStrunk:
We don't vendor the modules, but some golang builder containers assume
we do. This forces vendor directories to be ignored.

**Related issues:**
- openshift/release#20036


JohnStrunk:
/retitle Ensure golang modules are downloaded, not vendored

PR #92, opened 2021-09-30
Use ENV instead of ARG to override GOFLAGS

JohnStrunk:
#91 didn't work as expected because the env var from the base container didn't get overwritten. Using ENV instead of ARG does overwrite.

**Related issues:**


sallyom (approved):
/lgtm

Issue #105, opened 2021-11-11
FIPS support


JohnStrunk:
**Describe the feature you'd like to have.**
VolSync should be FIPS 140 compliant

**What is the value to the end user? (why is it a priority?)**

**How will we know we have a good solution? (acceptance criteria)**
- VolSync operator and associated containers should only use FIPS 140 approved algorithms
- When running on a FIPS enabled host, VolSync should use the proper, approved crypto libraries
- The above items should be automatically verified in CI

**Additional context**

JohnStrunk:
This was fixed by #107 
/close

PR #107, opened 2021-11-19
Update container builds for FIPS compatibility

JohnStrunk:
- Switches builder containers to ones that can use fips modules
- Enable CGO since that's required to call out to certified modules
- Add a command in the build process to ensure library hooks are in place

**Related issues:**
#105

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 # Build the manager binary
-FROM golang:1.17 as builder
+FROM registry.access.redhat.com/ubi8/go-toolset as builder
+USER root
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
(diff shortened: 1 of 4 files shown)


JohnStrunk:
/cc @tesshuflower 

Since I switched the rsync mover container from centos:8 --> ubi-minimal, the openshift ci image builds will fail until I can fix the image substitution. This is because centos uses `yum` whereas ubi-minimal uses `microdnf` for installing packages. CI is still substituting in the centos image even though it is changed in this PR.

openshift/release#23815 has the fix to the image substitution, but I'd like to be able to verify it before merging, so we need these changes in first so that the job can be rehearsed with the updated dockerfiles.

tesshuflower (commented):
/lgtm

PR #111, opened 2021-11-29
CLI: Migration create

v-harihar:
This sub-command should prepare a destination to receive incoming transfers from an external source.

- [ ] Saves relationship info
- [ ] Creates the Namespace if it doesn't exist
- [ ] Creates a PVC in the Namespace if it doesn't exist
- [ ] Creates an associated ReplicationDestination

### DIFF
--- a/kubectl-volsync/cmd/migration.go
+++ b/kubectl-volsync/cmd/migration.go
@@ -21,10 +21,69 @@ import (
 	"github.com/spf13/viper"
 	"k8s.io/kubectl/pkg/util/i18n"
 	"k8s.io/kubectl/pkg/util/templates"
+
+	volsyncv1alpha1 "github.com/backube/volsync/api/v1alpha1"
 )
 
 // MigrationRelationship defines the "type" of migration Relationships
-const MigrationRelationship RelationshipType = "migration"
+const MigrationRelationshipType RelationshipType = "migration"
+
+// migrationRelationship holds the config state for migration-type
+// relationships
+type migrationRelationship struct {
+	Relationship
+	data *migrationRelationshipData
+}
+
+// migrationRelationshipData is the state that will be saved to the
+// relationship config file
+type migrationRelationshipData struct {
+	Version     int
+	Destination *migrationRelationshipDestination
+}
+
+type migrationRelationshipDestination struct {
+	// Cluster context name
+	Cluster string
+	// Namespace on destination cluster
+	Namespace string
+	// Name of PVC being replicated
+	PVCName string
+	// Name of the ReplicationDestination object
+	RDName string
+	// Name of Secret holding SSH keys
+	SSHKeyName string
+	// Parameters for the ReplicationDestination
+	Destination volsyncv1alpha1.ReplicationDestinationRsyncSpec
+}
+
+func (mr *migrationRelationship) Save() error {
+	err := mr.SetData(mr.data)
+	if err != nil {
+		return err
+	}
+
+	if mr.data.Destination != nil && mr.data.Destination.Destination.Capacity != nil {
+		mr.Set("data.destination.destination.replicationdestinationvolumeoptions.capacity",
+			mr.data.Destination.Destination.Capacity.String())
+	}
+
+	return mr.Relationship.Save()
+}
+
+func newMigrationRelationship(cmd *cobra.Command) (*migrationRelationship, error) {
+	r, err := CreateRelationshipFromCommand(cmd, MigrationRelationshipType)
+	if err != nil {
+		return nil, err
+	}
+
+	return &migrationRelationship{
+		Relationship: *r,
+		data: &migrationRelationshipData{
+			Version: 1,
+		},
+	}, nil
+}
 
 // migrationCmd represents the migration command
 var migrationCmd = &cobra.Command{
--- a/kubectl-volsync/cmd/migration_create.go
+++ b/kubectl-volsync/cmd/migration_create.go
@@ -17,13 +17,56 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 package cmd
 
 import (
+	"context"
 	"fmt"
+	"time"
 
 	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	kerrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
 	"k8s.io/kubectl/pkg/util/i18n"
 	"k8s.io/kubectl/pkg/util/templates"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	volsyncv1alpha1 "github.com/backube/volsync/api/v1alpha1"
 )
+
+type migrationCreate struct {
+	// migration relationship object to be persisted to a config file
+	mr *migrationRelationship
+	// Cluster context name
+	Cluster string
+	// Namespace on destination cluster
+	Namespace string
+	// destinationPVC is a PVC to use as the transfer destination instead of
+	// automatically provisioning one. Either this field or both capacity and
+	// accessModes must be specified.
+	DestinationPVC string
+	// Name of the ReplicationDestination object
+	RDName string
+	// copyMethod describes how a point-in-time (PiT) image of the destination
+	// volume should be created
+	CopyMethod volsyncv1alpha1.CopyMethodType
+	// capacity is the size of the destination volume to create
+	Capacity *resource.Quantity
+	// storageClassName can be used to specify the StorageClass of the
+	// destination volume. If not set, the default StorageClass will be used.
+	StorageClassName *string
+	// AccessModes contains the desired access modes the volume should have
+	AccessModes []corev1.PersistentVolumeAccessMode
+	// serviceType determines the Service type that will be created for incoming
+	// SSH connections.
+	ServiceType *corev1.ServiceType
+	// client object to communicate with a cluster
+	clientObject client.Client
+	// PVC object associated with pvcName used to create destination object
+	PVC *corev1.PersistentVolumeClaim
+}
 
 // migrationCreateCmd represents the create command
 var migrationCreateCmd = &cobra.Command{
@@ -36,70 +79,321 @@ var migrationCreateCmd = &cobra.Command{
 	and it sets up an associated ReplicationDestination that will be configured
 	to accept incoming transfers via rsync over ssh.
 	`)),
-	RunE: doMigrationCreate,
-	Args: validateMigrationCreate,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		mc, err := newMigrationCreate(cmd)
+		if err != nil {
+			return err
+		}
+		return mc.Run(cmd.Context())
+	},
 }
 
 func init() {
+	initMigrationCreateCmd(migrationCreateCmd)
+}
+
+func initMigrationCreateCmd(migrationCreateCmd *cobra.Command) {
 	migrationCmd.AddCommand(migrationCreateCmd)
 
-	migrationCreateCmd.Flags().String("accessmodes", "", "AccessModes of the PVC to create")
-	migrationCreateCmd.Flags().String("capacity", "", "capacity of the PVC to create")
+	migrationCreateCmd.Flags().String("accessmodes", "ReadWriteOnce",
+		"accessMode of the PVC to create. viz: ReadWriteOnce, ReadOnlyMany, ReadWriteMany, ReadWriteOncePod")
+	migrationCreateCmd.Flags().String("capacity", "", "size of the PVC to create (ex: 100Mi, 10Gi, 2Ti)")
 	migrationCreateCmd.Flags().String("pvcname", "", "name of the PVC to create or use: [context/]namespace/name")
 	cobra.CheckErr(migrationCreateCmd.MarkFlagRequired("pvcname"))
 	migrationCreateCmd.Flags().String("storageclass", "", "StorageClass name for the PVC")
+	migrationCreateCmd.Flags().String("servicetype", "LoadBalancer",
+		"Service Type or ingress methods for a service. viz: ClusterIP, LoadBalancer")
 }
 
-func validateMigrationCreate(cmd *cobra.Command, args []string) error {
-	// If specified, the PVC's capacity must parse to a valid resource.Quantity
-	capacity, err := cmd.Flags().GetString("capacity")
+func newMigrationCreate(cmd *cobra.Command) (*migrationCreate, error) {
+	mc := &migrationCreate{}
+	// build struct migrationRelationship from cmd line args
+	mr, err := newMigrationRelationship(cmd)
+	if err != nil {
+		return nil, err
+	}
+	mc.mr = mr
+
+	if err = mc.parseCLI(cmd); err != nil {
+		return nil, err
+	}
+
+	return mc, nil
+}
+
+//nolint:funlen
+func (mc *migrationCreate) parseCLI(cmd *cobra.Command) error {
+	pvcname, err := cmd.Flags().GetString("pvcname")
+	if err != nil || pvcname == "" {
+		return fmt.Errorf("failed to fetch the pvcname, err = %w", err)
+	}
+	xcr, err := ParseXClusterName(pvcname)
+	if err != nil {
+		return fmt.Errorf("failed to parse cluster name from pvcname, err = %w", err)
+	}
+	mc.DestinationPVC = xcr.Name
+	mc.Namespace = xcr.Namespace
+	mc.Cluster = xcr.Cluster
+
+	sCapacity, err := cmd.Flags().GetString("capacity")
 	if err != nil {
 		return err
 	}
-	if len(capacity) > 0 {
-		if _, err := resource.ParseQuantity(capacity); err != nil {
+	if len(sCapacity) > 0 {
+		capacity, err := resource.ParseQuantity(sCapacity)
+		if err != nil {
 			return fmt.Errorf("capacity must be a valid resource.Quantity: %w", err)
 		}
+		mc.Capacity = &capacity
 	}
-	// The PVC name must be specified, and it needs to be in the right format
-	pvcname, err := cmd.Flags().GetString("pvcname")
+
+	accessMode, err := cmd.Flags().GetString("accessmodes")
+	if err != nil {
+		return fmt.Errorf("failed to fetch access mode, %w", err)
+	}
+
+	if corev1.PersistentVolumeAccessMode(accessMode) != corev1.ReadWriteOnce &&
+		corev1.PersistentVolumeAccessMode(accessMode) != corev1.ReadOnlyMany &&
+		corev1.PersistentVolumeAccessMode(accessMode) != corev1.ReadWriteMany &&
+		corev1.PersistentVolumeAccessMode(accessMode) != corev1.ReadWriteOncePod {
+		return fmt.Errorf("unsupported access mode: %v", accessMode)
+	}
+	accessModes := []corev1.PersistentVolumeAccessMode{corev1.PersistentVolumeAccessMode(accessMode)}
+	mc.AccessModes = accessModes
+
+	storageClass, err := cmd.Flags().GetString("storageclass")
+	if err != nil {
+		return fmt.Errorf("failed to fetch storageClass, %w", err)
+	}
+	if storageClass == "" {
+		mc.StorageClassName = nil
+	} else {
+		mc.StorageClassName = &storageClass
+	}
+	// For migration use case, only Copymethod="Direct" is supported
+	mc.CopyMethod = volsyncv1alpha1.CopyMethodDirect
+
+	serviceType, err := cmd.Flags().GetString("servicetype")
+	if err != nil {
+		return fmt.Errorf("please provide service type, err = %w", err)
+	}
+
+	if corev1.ServiceType(serviceType) != corev1.ServiceTypeClusterIP &&
+		corev1.ServiceType(serviceType) != corev1.ServiceTypeLoadBalancer {
+		return fmt.Errorf("unsupported service type: %v", corev1.ServiceType(serviceType))
+	}
+	mc.ServiceType = (*corev1.ServiceType)(&serviceType)
+	mc.RDName = mc.Namespace + "-" + mc.DestinationPVC + "-migration-dest"
+
+	return nil
+}
+
+//nolint:funlen
+func (mc *migrationCreate) newMigrationRelationshipDestination() (
+	*migrationRelationshipDestination, error) {
+	mrd := &migrationRelationshipDestination{}
+
+	// Assign the values from migrationCreate built after parsing cmd args
+	mrd.RDName = mc.RDName
+	mrd.PVCName = mc.DestinationPVC
+	mrd.Namespace = mc.Namespace
+	mrd.Cluster = mc.Cluster
+
+	if mc.PVC == nil {
+		if mc.Capacity == nil {
+			return nil, fmt.Errorf("capacity arg must be provided")
+		}
+	}
+
+	mrd.Destination.DestinationPVC = &mc.DestinationPVC
+	mrd.Destination.ServiceType = mc.ServiceType
+
+	return mrd, nil
+}
+
+func (mc *migrationCreate) Run(ctx context.Context) error {
+	k8sClient, err := newClient(mc.Cluster)
 	if err != nil {
 		return err
 	}
-	if _, err := ParseXClusterName(pvcname); err != nil {
+	mc.clientObject = k8sClient
+
+	// Get the pvc from cluster
+	mc.PVC, err = mc.getDestinationPVC(ctx)
+	if err != nil {
 		return err
+	}
+
+	// Build struct migrationRelationshipDestination from struct migrationCreate
+	mc.mr.data.Destination, err = mc.newMigrationRelationshipDestination()
+	if err != nil {
+		return err
+	}
+	// Creates the Namespace if it doesn't exist
+	_, err = mc.ensureNamespace(ctx)
+	if err != nil {
+		return err
+	}
+	// Creates the PVC if it doesn't exist
+	_, err = mc.ensureDestPVC(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Creates the RD if it doesn't exist
+	_, err = mc.ensureReplicationDestination(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Wait for ReplicationDestination to post address, sshkeys
+	err = mc.waitForRDStatus(ctx)
+	if err != nil {
+		return err
+	}
+	// Save the destination details into relationship file
+	if err = mc.mr.Save(); err != nil {
+		return fmt.Errorf("unable to save relationship configuration: %w", err)
 	}
 	return nil
 }
 
-func doMigrationCreate(cmd *cobra.Command, args []string) error {
-	// Create the empty relationship
-	configDir, err := cmd.Flags().GetString("config-dir")
-	if err != nil {
-		return err
+func (mc *migrationCreate) ensureNamespace(ctx context.Context) (*corev1.Namespace, error) {
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: mc.Namespace,
+		},
 	}
-	rName, err := cmd.Flags().GetString("relationship")
-	if err != nil {
-		return err
+	if err := mc.clientObject.Create(ctx, ns); err != nil {
+		if kerrs.IsAlreadyExists(err) {
+			klog.Infof("Namespace: \"%s\" is found, proceeding with the same",
+				mc.Namespace)
+			return ns, nil
+		}
+		return nil, err
 	}
-	relation, err := createRelationship(configDir, rName, MigrationRelationship)
-	if err != nil {
-		return err
+	klog.Infof("Created Destination Namespace: \"%s\" in Cluster: \"%s\"", ns.Name, ns.ClusterName)
+
+	return ns, nil
+}
+
+func (mc *migrationCreate) ensureDestPVC(ctx context.Context) (*corev1.PersistentVolumeClaim, error) {
+	if mc.PVC == nil {
+		PVC, err := mc.createDestinationPVC(ctx)
+		if err != nil {
+			return nil, err
+		}
+		mc.PVC = PVC
+	} else {
+		klog.Infof("Destination PVC: \"%s\" is found in Namespace: \"%s\" and is used to create replication destination",
+			mc.PVC.Name, mc.PVC.Namespace)
 	}
 
-	// Insert information into the relationship & save it
-	cap, err := cmd.Flags().GetString("capacity")
+	return mc.PVC, nil
+}
+
+func (mc *migrationCreate) createDestinationPVC(ctx context.Context) (*corev1.PersistentVolumeClaim, error) {
+	destPVC := &corev1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      mc.DestinationPVC,
+			Namespace: mc.Namespace,
+		},
+		Spec: corev1.PersistentVolumeClaimSpec{
+			AccessModes:      mc.AccessModes,
+			StorageClassName: mc.StorageClassName,
+			Resources: corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{
+					corev1.ResourceStorage: *mc.Capacity,
+				},
+			},
+		},
+	}
+
+	if err := mc.clientObject.Create(ctx, destPVC); err != nil {
+		return nil, err
+	}
+
+	klog.Infof("Created Destination PVC: \"%s\" in NameSpace: \"%s\" and Cluster: \"%s\" ",
+		destPVC.Name, destPVC.Namespace, destPVC.ClusterName)
+
+	return destPVC, nil
+}
+
+func (mc *migrationCreate) getDestinationPVC(ctx context.Context) (*corev1.PersistentVolumeClaim, error) {
+	destPVC := &corev1.PersistentVolumeClaim{}
+	pvcInfo := types.NamespacedName{
+		Namespace: mc.Namespace,
+		Name:      mc.DestinationPVC,
+	}
+	err := mc.clientObject.Get(ctx, pvcInfo, destPVC)
 	if err != nil {
-		return err
+		if client.IgnoreNotFound(err) == nil {
+			klog.Infof("pvc: \"%s\" not found, creating the same", mc.DestinationPVC)
+			return nil, nil
+		}
+		return nil, err
 	}
-	relation.Set("capacity", cap)
+	return destPVC, nil
+}
 
-	if err = relation.Save(); err != nil {
-		return fmt.Errorf("unable to save relationship configuration: %w", err)
+func (mc *migrationCreate) ensureReplicationDestination(ctx context.Context) (
+	*volsyncv1alpha1.ReplicationDestination, error) {
+	mrd := mc.mr.data.Destination
+	rd := &volsyncv1alpha1.ReplicationDestination{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      mrd.RDName,
+			Namespace: mrd.Namespace,
+		},
+		Spec: volsyncv1alpha1.ReplicationDestinationSpec{
+			Rsync: &volsyncv1alpha1.ReplicationDestinationRsyncSpec{
+				ReplicationDestinationVolumeOptions: volsyncv1alpha1.ReplicationDestinationVolumeOptions{
+					DestinationPVC: mrd.Destination.DestinationPVC,
+				},
+				ServiceType: mrd.Destination.ServiceType,
+			},
+		},
 	}
+	if err := mc.clientObject.Create(ctx, rd); err != nil {
+		return nil, err
+	}
+	klog.Infof("Created ReplicationDestination: \"%s\" in Namespace: \"%s\" and Cluster: \"%s\"",
+		rd.Name, rd.Namespace, rd.ClusterName)
 
-	// TODO: Set up objects on the cluster
-	_, _ = newClient("")
+	return rd, nil
+}
+
+func (mc *migrationCreate) waitForRDStatus(ctx context.Context) error {
+	mrd := mc.mr.data.Destination
+	// wait for migrationdestination to become ready
+	nsName := types.NamespacedName{
+		Namespace: mrd.Namespace,
+		Name:      mrd.RDName,
+	}
+	rd := &volsyncv1alpha1.ReplicationDestination{}
+	err := wait.PollImmediate(5*time.Second, 2*time.Minute, func() (bool, error) {
+		err := mc.clientObject.Get(ctx, nsName, rd)
+		if err != nil {
+			return false, err
+		}
+		if rd.Status == nil || rd.Status.Rsync == nil {
+			return false, nil
+		}
+		if rd.Status.Rsync.Address == nil {
+			klog.V(2).Infof("Waiting for MigrationDestination %s RSync address to populate", rd.Name)
+			return false, nil
+		}
+
+		if rd.Status.Rsync.SSHKeys == nil {
+			klog.V(2).Infof("Waiting for MigrationDestination %s RSync sshkeys to populate", rd.Name)
+			return false, nil
+		}
+
+		klog.V(2).Infof("Found MigrationDestination RSync Address: %s", *rd.Status.Rsync.Address)
+		return true, nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to fetch rd status: %w,", err)
+	}
 
 	return nil
 }
--- a/kubectl-volsync/cmd/migration_create_test.go
+++ b/kubectl-volsync/cmd/migration_create_test.go
@@ -1,0 +1,193 @@
+package cmd
+
+import (
+	"context"
+	"io/ioutil"
+	"os"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	krand "k8s.io/apimachinery/pkg/util/rand"
+
+	volsyncv1alpha1 "github.com/backube/volsync/api/v1alpha1"
+)
+
+var _ = Describe("migration", func() {
+	var (
+		ns               *corev1.Namespace
+		cmd              *cobra.Command
+		mc               *migrationCreate
+		migrationCmdArgs = make(map[string]string)
+		dirname          string
+	)
+
+	BeforeEach(func() {
+		ns = &corev1.Namespace{}
+		cmd = &cobra.Command{}
+		mc = &migrationCreate{}
+		var err error
+
+		migrationCmdArgs = map[string]string{
+			"capacity": "2Gi",
+			"pvcname":  "dest/volsync",
+		}
+
+		initMigrationCreateCmd(cmd)
+		cmd.Flags().String("relationship", "test", "")
+
+		dirname, err = ioutil.TempDir("", "relation")
+		Expect(err).NotTo(HaveOccurred())
+		cmd.Flags().String("config-dir", dirname, "")
+
+		mr, err := newMigrationRelationship(cmd)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(mr).ToNot(BeNil())
+		mc.mr = mr
+
+		err = migrationCmdArgsSet(cmd, migrationCmdArgs)
+		Expect(err).ToNot(HaveOccurred())
+
+		err = mc.parseCLI(cmd)
+		Expect(err).NotTo(HaveOccurred())
+
+		mc.clientObject = k8sClient
+		mc.Namespace = "foo-" + krand.String(5)
+		// create namespace
+		ns, err = mc.ensureNamespace(context.Background())
+		Expect(err).ToNot(HaveOccurred())
+		Expect(ns).NotTo(BeNil())
+
+		ns = &corev1.Namespace{}
+		Expect(k8sClient.Get(context.Background(), types.NamespacedName{Name: mc.Namespace}, ns)).To(Succeed())
+
+	})
+
+	AfterEach(func() {
+		if ns != nil {
+			Expect(k8sClient.Delete(context.Background(), ns)).To(Succeed())
+		}
+
+		os.RemoveAll(dirname)
+	})
+
+	It("Verify migration create arguments: fails with missing pvcname arg, with no pre-existing PVC", func() {
+		err := cmd.Flags().Set("pvcname", "")
+		Expect(err).NotTo(HaveOccurred())
+		err = mc.parseCLI(cmd)
+		Expect(err).To(HaveOccurred())
+	})
+
+	It("Verify migration create arguments: works with minimum set of arguments, with no preexsting PVC ", func() {
+		err := mc.parseCLI(cmd)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("Verify migration create arguments: fails with wrong accessMode arg value", func() {
+		err := cmd.Flags().Set("accessmodes", "Read-Write-Once")
+		Expect(err).NotTo(HaveOccurred())
+		err = mc.parseCLI(cmd)
+		Expect(err).To(HaveOccurred())
+	})
+
+	It("Verify migration create arguments: fails with wrong capacity arg value", func() {
+		err := cmd.Flags().Set("capacity", "2GB")
+		Expect(err).NotTo(HaveOccurred())
+		err = mc.parseCLI(cmd)
+		Expect(err).To(HaveOccurred())
+	})
+
+	It("Ensure namespace creation", func() {
+		ns = &corev1.Namespace{}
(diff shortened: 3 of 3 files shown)


kubectl-volsync/cmd/migration.go
```diff
@@ -19,12 +19,86 @@ package cmd
...
+	PVC *v1.PersistentVolumeClaim
+	// Name of the migrationDestination object
+	MDName string
+	// client object associated with a cluster
+	clientObject client.Client
+	// Name of Secret holding SSH keys
+	SSHKeyName string
+	// Parameters for the migrationDestination
+	Destination volsyncv1alpha1.ReplicationDestinationRsyncSpec
+}
```
JohnStrunk, line 73: The PVC object and the client.Client shouldn't be serialized to the config file.
- The *name* of the PVC can be saved, but it's probably easiest to use `Destination.DestinationPVC` for that.
- The client.Client can be reconstructed from the Cluster context name.

kubectl-volsync/cmd/migration.go
```diff
@@ -19,12 +19,86 @@ package cmd
...
+	Source      *migrationRelationshipSource
+	Destination *migrationRelationshipDestination
+}
+
+type migrationRelationshipSource struct {
+	// Volume to be migrated
+	Volume string
+	// Total volume size
+	Size *resource.Quantity
+}
```
JohnStrunk, line 54: The external -> kubernetes workflow doesn't really have a "source" for the relationship.
The Size parameter can be stored in Destination.Capacity

JohnStrunk (changes requested):
I think a lot of the code is there, it just needs a bit of rearrangement.

kubectl-volsync/cmd/migration.go
v-harihar: Sure. I will remove `PVC` object and  `client.Client ` from `migrationRelationshipDestination` struct and keep them as global vars.

kubectl-volsync/cmd/migration.go
v-harihar: Shall we emphasize the capacity description as it is the size of the volume to be migrated? something like

`migrationCreateCmd.Flags().String("capacity", "", "Size of the standalone volume to be migrated. The PVC that is to be created is of this size. format ex: 10Gi")`

kubectl-volsync/cmd/migration.go
```diff
@@ -19,12 +19,86 @@ package cmd
...
+
+func (mr *migrationRelationship) Save() error {
+	mr.Set("data", mr.data)
+	// resource.Quantity doesn't properly encode, so we need to do it manually
+	/*
+		if mr.data.Source != nil && mr.data.Source.Source.Capacity != nil {
+			mr.Set("data.source.source.capacity", mr.data.Source.Source.Capacity.String())
+		}
+	*/
```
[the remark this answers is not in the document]
v-harihar: done

kubectl-volsync/cmd/migration_create.go
```diff
@@ -36,18 +48,29 @@ var migrationCreateCmd = &cobra.Command{
...
 	Args: validateMigrationCreate,
 }
 
 func init() {
 	migrationCmd.AddCommand(migrationCreateCmd)
 
-	migrationCreateCmd.Flags().String("accessmodes", "", "AccessModes of the PVC to create")
-	migrationCreateCmd.Flags().String("capacity", "", "capacity of the PVC to create")
+	migrationCreateCmd.Flags().String("copymethod", "",
+		"Copy method used to copy the data from source to destination PVC")
```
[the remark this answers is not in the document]
v-harihar: In `copymethod = Direct`, we just copy data to destination PVC and don't create a snapshot of the PVC?

kubectl-volsync/cmd/migration_create.go
```diff
@@ -36,18 +48,29 @@ var migrationCreateCmd = &cobra.Command{
...
+	migrationCreateCmd.Flags().String("accessmodes", "",
+		"accessMode of the PVC to create. viz: ReadWriteOnce, ReadOnlyMany, ReadWriteMany, ReadWriteOncePod")
+	migrationCreateCmd.Flags().String("capacity", "", "size of the PVC to create in Gi ex: 10Gi")
```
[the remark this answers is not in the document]
v-harihar: `access mode and capacity`  should be mandatory args?

If PVC pre-exists then capacity won't be a requirement, since we are using the pre-existing PVC.
Also, the access mode can default to `READWRITEONCE` if not provided

kubectl-volsync/cmd/migration_create.go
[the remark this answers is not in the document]
JohnStrunk: Allowing the user to specify that they want "Snapshot" as a copymethod is fine (Clone isn't valid on the destination); there are workflows where it may make sense. I believe Direct will be the majority, so we should default it it.

kubectl-volsync/cmd/migration.go
JohnStrunk: globals make it hard to test... ideally pass then as params or add them to the migrationCreate struct

kubectl-volsync/cmd/migration.go
JohnStrunk: I'd like to keep it clear that the capacity is the PVC size. While it's true that the copy will fail if the PVC is too small, it needs to be really obvious that's what they are setting w/ this option. There are valid reasons to both oversize (for growth) or undersize (not migrating everything).

kubectl-volsync/cmd/migration_create.go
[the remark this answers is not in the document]
JohnStrunk: good catch. accessMode and capacity are only needed if PVC doesn't exist.

I'm ok w/ defaulting to RWO.

kubectl-volsync/cmd/migration.go
v-harihar: makes sense

kubectl-volsync/cmd/migration.go
v-harihar: done

kubectl-volsync/cmd/migration.go
```diff
@@ -19,12 +19,86 @@ package cmd
...
+// migrationRelationship holds the config state for migration-type
+// relationships
+type migrationRelationship struct {
+	Relationship
+	data migrationRelationshipData
+}
+
+// migrationRelationshipData is the state that will be saved to the
+// relationship config file
+type migrationRelationshipData struct {
```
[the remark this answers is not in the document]
v-harihar: done

kubectl-volsync/cmd/migration_create.go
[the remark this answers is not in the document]
v-harihar: sure

kubectl-volsync/cmd/migration_create.go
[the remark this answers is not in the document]
v-harihar: done

kubectl-volsync/cmd/migration.go
```diff
@@ -21,10 +21,75 @@ import (
...
+	// Name of Secret holding SSH keys
+	SSHKeyName string
+	// Parameters for the migrationDestination
+	Destination volsyncv1alpha1.ReplicationDestinationRsyncSpec
+}
+
+func (mr *migrationRelationship) Save() error {
+	mr.Set("data", mr.data)
+	// resource.Quantity doesn't properly encode, so we need to do it manually
+	if mr.data.Destination != nil && mr.data.Destination.Destination.Capacity != nil {
```
JohnStrunk, line 63: The `mr.Set()` on L61 will recursively put the mr.data into the Viper instance. The individual lower-level Set calls are only necessary for structs that don't serialize correctly (e.g., `resource.Capacity`).
As an example:
https://github.com/JohnStrunk/volsync/blob/2d0e8e6c374f5364c359def47b6be8dd6ee31bad/kubectl-volsync/cmd/replication.go#L72-L82
The set call encodes all of `.data`, but since Capacity doesn't encode correctly for the source and destination, those are set specifically.

Note that resource.Capacity types do decode correctly, so no special handling is needed in the load function.

kubectl-volsync/cmd/migration_create.go
```diff
@@ -36,18 +54,29 @@ var migrationCreateCmd = &cobra.Command{
...
 func init() {
 	migrationCmd.AddCommand(migrationCreateCmd)
 
-	migrationCreateCmd.Flags().String("accessmodes", "", "AccessModes of the PVC to create")
-	migrationCreateCmd.Flags().String("capacity", "", "capacity of the PVC to create")
+	migrationCreateCmd.Flags().String("copymethod", "Direct",
+		"Copy method used to copy the data from source to destination PVC viz: Direct, Snapshot")
+	migrationCreateCmd.Flags().String("accessmodes", "ReadWriteOnce",
+		"accessMode of the PVC to create. viz: ReadWriteOnce, ReadOnlyMany, ReadWriteMany, ReadWriteOncePod")
+	migrationCreateCmd.Flags().String("capacity", "", "size of the PVC to create in Gi ex: 10Gi")
```
JohnStrunk, line 73:
```suggestion
	migrationCreateCmd.Flags().String("capacity", "", "size of the PVC to create (ex: 10Gi)")
```
Other units are allowed (100Mi, 2Ti, etc)

kubectl-volsync/cmd/migration_create.go
```diff
@@ -18,13 +18,31 @@ package cmd
...
+	volsyncv1alpha1 "github.com/backube/volsync/api/v1alpha1"
 )
 
+type migrationCreate struct {
+	cobra.Command
+	// client object associated with a cluster
+	clientObject client.Client
+	// PVC object associated with pvcName
+	PVC *v1.PersistentVolumeClaim
+}
```
JohnStrunk, line 44: This can be the struct that all the "create" helper functions belong to.
We can also remove the Command my parsing the CLI in the newMigrationCreate() function
```go
type migrationCreate struct {
	rel *migrationRelationship
	// client object associated with a cluster
	clientObject client.Client
	// PVC object associated with pvcName
	PVC *v1.PersistentVolumeClaim
	// CLI options
	copyMethod volsyncv1alpha1.CopyMethodType
	//etc...
}
```

RunE below can be:
```go
RunE: func(cmd *cobra.Command, args []string) error {
	rsd, err := newMigrationCreate(cmd)
		if err != nil {
			return err
		}
		return rsd.Run(cmd.Context())
}

newMigrationCreate is then the only function that has to parse flag strings and validate them

File: kubectl-volsync/cmd/migration.go (hunk 7 above)
Comment (v-harihar) [reply kubectl-volsync/cmd/migration.go:63]: Yeah. I tried with the below code (pre-existing)
```
func (mr *migrationRelationship) Save() error {
	mr.Set("data", mr.data)
	// resource.Quantity doesn't properly encode, so we need to do it manually
	if mr.data.Destination != nil && mr.data.Destination.Destination.Capacity != nil {
		mr.Set("data.destination.spec.Capacity", mr.data.Destination.Destination.Capacity.String())
```
that resulted in below o/p
**o/p 1:**
```
# cat /root/.volsync/v3.yaml 
data:
  destination:
    spec:
      capacity: 2Gi
id: f77a765f-f25e-4052-9678-06f7a4e66cfa
type: migration
```

If I only keep `mr.Set("data", mr.data)` and remove `mr.Set("data.destination.spec.Capacity", mr.data.Destination.Destination.Capacity.String())`, relationship data will be saved but capacity gets written in f`ormat: BinarySI`

o/p2:
```
data:
  version: 1
  destination:
    cluster: ""
    namespace: dest
    pvcname: volsync
    mdname: dest-volsync-migration-dest
    sshkeyname: ""
    destination:
      replicationdestinationvolumeoptions:
        copymethod: Direct
        capacity:
          format: BinarySI
        storageclassname: csi-hostpath-sc
        accessmodes:
        - ReadWriteOnce
        volumesnapshotclassname: null
        destinationpvc: null
      sshkeys: volsync-rsync-dest-src-dest-volsync-migration-dest
      servicetype: ClusterIP
      address: 10.96.188.200
      port: null
      path: null
      sshuser: null
id: 443ab627-f55f-40dd-9fa0-c108195a0f90
type: migration
```
Then I tried with ugly-looking multiple mr.set ()

**Observation:**
> mr.Set("data.destination.spec.Capacity", mr.data.Destination.Destination.Capacity.String()) 
   The above statement will override `mr.Set("data", mr.data)` and just writes `Capacity` under destination just like above **o/p 1**

File: kubectl-volsync/cmd/migration_create.go (hunk 8 above)
Comment (v-harihar) [reply kubectl-volsync/cmd/migration_create.go:73]: done

File: kubectl-volsync/cmd/migration_create.go (hunk 9 above)
Comment (v-harihar) [reply kubectl-volsync/cmd/migration_create.go:44]: IMO, `rsd.Run()` part is not required as we don't do anything other than create a namespace, pvc, destination in case of `migration-create`. 

So RunE can be simple `newMigrationCreate()` function like
`RunE: newMigrationCreate`

Also, I feel structs `migrationCreate` and `migrationRelationship` can be kept separate (as in this code).
Reason: Whatever objects we create will be stored as part of `migrationCreate` and not persisted. This is true in the case of `migration-rsync` or `migration-delete` as well. On the fly data only be made part of this struct.

Does linking `relationship` with `migrationCreate` struct yield any special purpose? Please correct me if I am wrong

Comment (JohnStrunk) [reply kubectl-volsync/cmd/migration_create.go:44]: Yes, you could put the construction of the migrationCreate object and its usage in the same function. I'd expect that to make writing the unit tests tests more difficult.
With the suggested structure, they are separated, so you could construct the instance manually, then `Run` it.

The reason I suggested putting `migrationCreate` and `migrationRelationship` together is because all the functions that will implement the `create` command are going to be methods on `migrationCreate`... just having regular, non-member functions ends up polluting the namespace since all commands are in the same golang package.

For example, you have:
```go
func createDestinationPVC(mc *migrationCreate, mrd *migrationRelationshipDestination) (*v1.PersistentVolumeClaim, error) { }
```

This needs to be
```go
func (mc *migrationCreate) createDestinationPVC() (*v1.PersistentVolumeClaim, error) { }
```
so that the **replication** create could also have a similar function.
*As an aside-- if we end up repeating a bunch of code, we'll find a way to pull these out of the individual commands and share the implementations. But let's wait until there's demonstrable re-use first.*

Since `createDestinationPVC` is a method on `migrationCreate`, if now has access to those members via `mc`. If `migrationRelationship` is a member of that struct, you now have access to it as well w/o passing it around explicitly. I don't think there are many cases where you'd want one w/o the other.
- `migrationCreate` has all the temporary context for the action you're trying to perform (create the relationship)
- `migrationRelationship` has all the context that needs to be carried across all the actions (create, sync, delete)

v-harihar: Makes sense.

kubectl-volsync/cmd/migration.go
JohnStrunk: sure enough. Turns out Viper doesn't actually flatten the config until it's written, so the initial set establishes a value for "data", but not the subkeys. So when we set("data.some.path"), that destroys the initial value for "data".
It looks possible to get the intended behavior by serializing then deserializing between the initial set and the 2nd one, but that's really ugly.  I'll see what else I can come up with.

kubectl-volsync/cmd/migration_create.go
```diff
@@ -72,34 +104,329 @@ func validateMigrationCreate(cmd *cobra.Command, args []string) error {
...
+		return fmt.Errorf("failed to fetch the pvcname, err = %w", err)
+	}
+	xcr, err := ParseXClusterName(pvcname)
+	if err != nil {
+		return fmt.Errorf("failed to parse cluster name from pvcname, err = %w", err)
+	}
+	mrd.PVCName = xcr.Name
+	mrd.Namespace = xcr.Namespace
+
+	mc.clientObject, err = newClient(mrd.Cluster)
```
JohnStrunk, line 151: Creating the client in the middle of things makes it hard to test. For example, the test framework starts an API server, but to access it, you'll have to use the provided client.Client, not one that is constructed from the provided cluster context name.

I'd suggest refactoring into high level steps of:
- parse CLI
- get client
- validate CLI (args relative to each other and the current cluster state)
- update cluster state

These can be in the Run() function and may not get unit tested, but that should be ok w/ proper structure since it's basically straight line code.
The steps can each be unit tested individually.

kubectl-volsync/cmd/migration_create.go
```diff
@@ -72,34 +104,329 @@ func validateMigrationCreate(cmd *cobra.Command, args []string) error {
...
+	if err := mc.clientObject.Create(ctx, destPVC); err != nil {
+		return nil, err
+	}
+	mrd.Destination.StorageClassName = destPVC.Spec.StorageClassName
+	klog.Infof("Created destination PVC: \"%s\"", mrd.PVCName)
+
+	return destPVC, nil
+}
+
+func (mc *migrationCreate) deleteDestinationPVC(ctx context.Context) error {
```
JohnStrunk, line 297: It looks like this is only here for the tests?
The tests should be structured to clean themselves up. Take a look at
https://github.com/backube/volsync/blob/main/controllers/mover/restic/restic_test.go#L220-L228
and
https://github.com/backube/volsync/blob/main/controllers/mover/restic/restic_test.go#L272-L275
that set up and clean up from each test. The setup creates a new Namespace in which the test can run, and the cleanup deletes that NS, causing all the (namespaced) objects created during the test to be deleted.

kubectl-volsync/cmd/migration.go
JohnStrunk: Solution for this is in #116

kubectl-volsync/cmd/migration_create_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+	It("Verify migration create arguments: works with minimum set of arguments, with no preexsting PVC ", func() {
+		cmd := &cobra.Command{}
+		cmd.Flags().String("pvcname", "dest/volsync", "")
+		cmd.Flags().String("capacity", "2Gi", "")
+		cmd.Flags().String("relationship", "test", "")
+		cmd.Flags().String("servicetype", "ClusterIP", "")
+		cmd.Flags().String("config-dir", "", "")
+		cmd.Flags().String("copymethod", "Direct", "")
+		cmd.Flags().String("accessmodes", "ReadWriteOnce", "")
```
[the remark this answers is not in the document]
v-harihar: I used to set flags using `cmd.Flags().Set()` but it was unnecessarily showing the help command when TC gets executed then I removed the `set` part and kept it like above.

Sure, I will do as you suggested.

kubectl-volsync/cmd/migration.go
```diff
@@ -64,6 +64,7 @@ func (mr *migrationRelationship) Save() error {
 	// If we set capacity explicitly, it will override mr.set("data", mr.data)
 	// behavior and only capacity will be set. So resetting every other fields
 	// explicitly.
+
```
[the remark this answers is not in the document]
v-harihar: Yeah. We can check and create the `/root/.volsync` dir before the create starts.

@JohnStrunk please suggest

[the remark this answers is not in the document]
JohnStrunk: #122

kubectl-volsync/cmd/migration_create.go
```diff
@@ -72,34 +108,294 @@ func validateMigrationCreate(cmd *cobra.Command, args []string) error {
...
+				AccessModes:      mrd.Destination.AccessModes,
+				StorageClassName: mrd.Destination.StorageClassName,
+				Resources: v1.ResourceRequirements{
+					Requests: v1.ResourceList{
+						v1.ResourceStorage: *mrd.Destination.Capacity,
+					},
+				},
+			},
+		}
+	}
```
JohnStrunk, line 304: A couple things about this block...

While we're not making the distinction (currently) between an empty StorageClassName (=="") and a nil one (==nil) on the cli, they have different meanings to Kubernetes. Nil means "use the default", while empty means "use one with an empty name". When we parse the cli, we should put the parameter into a `*string` so that we can preserve that distinction.

After parsing the cli arg into the `*string`, the duplication here can be removed.

kubectl-volsync/cmd/migration_create.go
```diff
@@ -72,34 +108,294 @@ func validateMigrationCreate(cmd *cobra.Command, args []string) error {
...
+	if err == nil {
+		return destPVC
+	}
+
+	return nil
+}
+
+func (mc *migrationCreate) createDestination(ctx context.Context) error {
+	mrd := mc.mr.data.Destination
+	rsyncSpec := &volsyncv1alpha1.ReplicationDestinationRsyncSpec{
```
JohnStrunk, line 331: My expectation had been that `data.Destination` was just going to get directly assigned to `rsyncSpec` as opposed to only some of the fields. I think a straight assignment would probably make it a bit more obvious elsewhere in the code (e.g. newMigrationRelationshipDestination) how the values were going to be used.

kubectl-volsync/cmd/migration_create.go
```diff
@@ -72,34 +108,294 @@ func validateMigrationCreate(cmd *cobra.Command, args []string) error {
...
+
+		klog.Infof("Found MigrationDestination RSync Address: %s", *rd.Status.Rsync.Address)
+		return true, nil
+	})
+	if err != nil {
+		return err
+	}
+	mrd.Destination.Address = rd.Status.Rsync.Address
+	mrd.Destination.Port = rd.Status.Rsync.Port
+	mrd.Destination.SSHKeys = rd.Status.Rsync.SSHKeys
```
JohnStrunk, line 383: I would suggest fetching these from the cluster when we want to use them as opposed to saving them in the relationship file. Though waiting for them to populate (as above) is probably good.

kubectl-volsync/cmd/migration_test_helper.go
```diff
@@ -0,0 +1,106 @@
+package cmd
```
JohnStrunk, line 1: Code only used for/in tests needs to go in files named `*_test.go`

kubectl-volsync/cmd/migration_create.go
```diff
@@ -36,18 +56,34 @@ var migrationCreateCmd = &cobra.Command{
...
 	`)),
-	RunE: doMigrationCreate,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		mc, err := newMigrationCreate(cmd)
+		if err != nil {
+			return err
+		}
+		return mc.Run(cmd.Context())
+	},
 	Args: validateMigrationCreate,
```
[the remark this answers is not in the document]
v-harihar: Yes, Moving the validation part to `newMigrationCreate` and removing args.

kubectl-volsync/cmd/migration_create.go
```diff
@@ -72,34 +108,294 @@ func validateMigrationCreate(cmd *cobra.Command, args []string) error {
...
+	// build struct migrationRelationshipDestination from cmd line args
+	err = newMigrationRelationshipDestination(mc, cmd)
+	if err != nil {
+		return nil, err
+	}
+
+	return mc, nil
+}
+
+func parsePVCName(cmd *cobra.Command, mrd *migrationRelationshipDestination) error {
```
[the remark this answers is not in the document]
v-harihar: Please correct my understanding, once we parse the CLI/cmd args, the placeholder to store the values is under `migrationRelationshipDestination` that is `migrationCreate.migrationRelationship.migrationRelationshipData.migrationRelationshipDestination`.
Few members are direct members of `migrationRelationshipDestination` and few are part of `migrationRelationshipDestination.Destination`. After parsing, the entire `migrationRelationshipDestination` along with `migrationRelationshipDestination.Destination` will be built. Once we are ready with the structure then we can call `Run` just to do the actual work (not required to build `migrationRelationshipDestination.Destination` structure). 

If we push the building of  `migrationRelationshipDestination.Destination` in `Run`, then we again need to parse the CLI as we don't have any other fields to store the CLI values temporarily during `newMigrationCreate`.

With the current approach
1. parse and build entire `migrationRelationshipDestination` in `newMigrationCreate`
2. do the actual work as part of `Run`

> - If the parsing is isolated in the newMigrationCreate, you have the opportunity in your tests to manually build the migrationCreate struct via a simple assignment and not worry about the parsing logic (unless that's what you're testing).
> 
> - And if you're testing the parsing, you can build the cmd and call newMigrationCreate, then directly check the contents of the struct.

Yes, this is the approach I am following.

kubectl-volsync/cmd/migration_create.go
```diff
@@ -72,34 +108,294 @@ func validateMigrationCreate(cmd *cobra.Command, args []string) error {
...
-	rName, err := cmd.Flags().GetString("relationship")
+	mrd.PVCName = xcr.Name
+	mrd.Namespace = xcr.Namespace
+	mrd.Cluster = xcr.Cluster
+
+	return nil
+}
+
+//nolint:funlen
+func newMigrationRelationshipDestination(mc *migrationCreate, cmd *cobra.Command) error {
```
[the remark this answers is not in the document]
v-harihar: Yes, only parsing the args and filling up the `Destination` struct as explained in https://github.com/backube/volsync/pull/111#discussion_r781718917 . This is called by `newMigrationCreate()`. 

Am I missing anything?

kubectl-volsync/cmd/migration_create.go
v-harihar: sure. will set `StorageClassName = nil` if arg is coming empty

kubectl-volsync/cmd/migration_create.go
v-harihar: sure. I will assign rsyncSpec coming from top and not build it here

kubectl-volsync/cmd/migration_create.go
v-harihar: Okay, sure. I will remove this if not required to be saved as part of relationship file

kubectl-volsync/cmd/migration_test_helper.go
v-harihar: sure

kubectl-volsync/cmd/migration_create.go
```diff
@@ -72,34 +108,294 @@ func validateMigrationCreate(cmd *cobra.Command, args []string) error {
...
+		if err != nil || storageClass == "" {
+			klog.Infof("storage class not provided, binding to default storage class")
+		}
+		mrd.Destination.StorageClassName = &storageClass
+	} else {
+		mrd.Destination.Capacity = mc.PVC.Spec.Resources.Requests.Storage()
+		mrd.Destination.AccessModes = mc.PVC.Spec.AccessModes
+		mrd.Destination.StorageClassName = mc.PVC.Spec.StorageClassName
+		mrd.Cluster = mc.PVC.ClusterName
+	}
```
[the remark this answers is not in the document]
v-harihar: > to clarify, the capacity, accessmodes, and storageclass should only be set in the ReplicationDestination if DestinationPVC is not set

This is in the case of pre-existing PVC right? Are we not updating the relationship file with capacity, accessmodes and storageclass if PVC already present ?

kubectl-volsync/cmd/migration_create.go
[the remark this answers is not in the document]
JohnStrunk: The parsed flags go directly in `migrationCreate`, not in the migration relationship struct since they don't need to be preserved or used across the different sub-commands.
https://github.com/JohnStrunk/volsync/blob/cli-replication/kubectl-volsync/cmd/replication_setSource.go#L36-L38

`migrationCreate.Run()` then has the logic to take the parsed options and set the relationship structure to reflect what should be in the cluster. Then it can be `migrationRelationship.Save()`-ed and applied to the cluster.

------

In #123 (cli replication), I'm experimenting with having an `Apply()` function that belongs to the replcationRelationship struct (and is therefore common across all of its sub-commands. Apply takes the relationship and updates the cluster accordingly. This allows all the sub-commands to just have logic in their `Run()` to update the relationship struct, then call Apply as the final step. I don't yet know how this will work out, but I'm hopeful.

kubectl-volsync/cmd/migration_create.go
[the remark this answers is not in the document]
JohnStrunk: If DestinationPVC is set, the operator will not create a PVC (it'll just use the DestinationPVC one). Capacity, accessmodes, and storageclass (in .spec.rsync) are only used by the operator when creating a PVC.

case 2 is important because if the operator creates the PVC, the PVC will be deleted when the ReplicationDestination object is deleted, and we don't want that to happen if we're using Direct/None.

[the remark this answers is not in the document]
v-harihar: Handling copymethod=Direct only for the migration use case. So considering case 1 and case 2  as per our discussion

kubectl-volsync/cmd/migration_create.go
```diff
@@ -36,70 +93,309 @@ var migrationCreateCmd = &cobra.Command{
...
+		// Default "AccessMode = ReadWriteOnce" if not provided
+		mrd.Destination.AccessModes = mc.AccessModes
+
+		if mc.StorageClassName == nil {
+			klog.Infof("Storage class is not specified as part of cmd arg, binding to default storage class")
+			mrd.Destination.StorageClassName = nil
+		} else {
+			mrd.Destination.StorageClassName = mc.StorageClassName
+		}
+
```
JohnStrunk, line 231: Since the operator isn't ever going to create the PVC (we're only supporting copymethod direct), we shouldn't be setting these fields in the ReplicationDestination--- just DestinationPVC.

kubectl-volsync/cmd/migration_create.go
```diff
@@ -36,70 +93,309 @@ var migrationCreateCmd = &cobra.Command{
...
+		}
+	} else {
+		klog.Infof("Destination PVC: \"%s\" is found in Namespace: \"%s\" and is used to create replication destination",
+			mc.PVC.Name, mc.PVC.Namespace)
+	}
+	// create migration destination
+	err = mc.createDestination(ctx)
 	if err != nil {
 		return err
 	}
```
JohnStrunk, line 278: I suggest a slightly different approach that's similar to how the operator itself works...
If you change the ordering:
- Namespace
- PVC
- ReplicationDestination

...you'll have the info available as you need it.

Then for the functions that do the work, think of them as "ensuring the state is a certain way". So consider:
```go
// Creates the Namespace if it doesn't exist
func (mc *migrationCreate) ensureNamespace (ctx context.Context) (*corev1.Namespace, error) {
...
}

// Creates the PVC if it doesn't exist
func (mc *migrationCreate) ensureDestPVC (ctx context.Context) (*corev1.PersistentVolumeClaim, error) {
...
}

// Creates the RD if it doesn't exist
func (mc *migrationCreate) ensureReplicationDestination (ctx context.Context) (*volsyncv1alpha1.ReplicationDestination, error) {
...
}
```

Then Run() becomes just calling them in sequence. And you know when writing, for example, ensureReplicationDestination that the PVC is there, and you have a pointer to it in mc.PVC.

kubectl-volsync/cmd/migration_create.go
```diff
@@ -17,14 +17,71 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 package cmd
 
 import (
+	"context"
 	"fmt"
+	"time"
 
 	"github.com/spf13/cobra"
+	v1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
v-harihar: done

kubectl-volsync/cmd/migration_create.go
```diff
@@ -17,14 +17,71 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
...
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
 	"k8s.io/kubectl/pkg/util/i18n"
 	"k8s.io/kubectl/pkg/util/templates"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/backube/volsync/api/v1alpha1"
```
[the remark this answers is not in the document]
v-harihar: My bad. That happened because of auto import.

kubectl-volsync/cmd/migration_create.go
```diff
@@ -215,15 +197,67 @@ func newMigrationRelationshipDestination(mc *migrationCreate, cmd *cobra.Command
...
+	mrd.Destination.ServiceType = mc.ServiceType
 
 	return nil
 }
 
 func (mc *migrationCreate) Run(ctx context.Context) error {
+	mrd := &migrationRelationshipDestination{}
+	mc.mr.data.Destination = mrd
+
+	k8sClient, err := newClient(mrd.Cluster)
```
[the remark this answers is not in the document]
v-harihar: My bad. I should have passed the cluster name and checked this in relationship file

kubectl-volsync/cmd/migration_create.go
```diff
@@ -17,14 +17,71 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
 	"k8s.io/kubectl/pkg/util/i18n"
 	"k8s.io/kubectl/pkg/util/templates"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/backube/volsync/api/v1alpha1"
+	volsyncv1alpha1 "github.com/backube/volsync/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
v-harihar: Yeah sure. Auto import mislead most of the time

kubectl-volsync/cmd/migration_create.go
v-harihar: sure

kubectl-volsync/cmd/migration_create.go
v-harihar: sure

kubectl-volsync/cmd/migration_create.go
```diff
@@ -17,14 +17,69 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
...
+	volsyncv1alpha1 "github.com/backube/volsync/api/v1alpha1"
 )
 
+type migrationCreate struct {
+	// migration relationship object to be persisted to a config file
+	mr *migrationRelationship
+	// contains replication destination metadata
+	replicationDestinationMetadata
+	// contains replication destination specs
+	replicationDestinationRsyncSpec
```
JohnStrunk, line 45: If these structs aren't used anywhere else, please just inline the fields.

kubectl-volsync/cmd/migration_create.go
```diff
@@ -36,70 +91,319 @@ var migrationCreateCmd = &cobra.Command{
...
+		if mc.StorageClassName == nil {
+			klog.Infof("Storage class is not specified as part of cmd arg, binding to default storage class")
+			mrd.Destination.StorageClassName = nil
+		} else {
+			mrd.Destination.StorageClassName = mc.StorageClassName
+		}
+
+		mrd.Destination.DestinationPVC = &mc.PVCName
+	} else {
+		mrd.Destination.DestinationPVC = &mc.PVC.Name
```
JohnStrunk, line 232: Since the operator is never creating the PVC, only mrd.Destination.DestinationPVC needs to be set. The other fields are only used if you need to create the PVC (and should be handled in ensureDestPVC).

kubectl-volsync/cmd/migration_create.go
```diff
@@ -36,70 +91,319 @@ var migrationCreateCmd = &cobra.Command{
...
+func (mc *migrationCreate) getDestinationPVC(ctx context.Context) *corev1.PersistentVolumeClaim {
+	destPVC := &corev1.PersistentVolumeClaim{}
+	pvcInfo := types.NamespacedName{
+		Namespace: mc.Namespace,
+		Name:      mc.PVCName,
+	}
+	err := mc.clientObject.Get(ctx, pvcInfo, destPVC)
+	if err != nil {
+		return nil
+	}
```
JohnStrunk, line 350: This treats all errors the same. "not found" should be different from "access denied".
not found --> we need to create it
other error --> problem we need to report

But I'm not convinced you actually need this fn at all. You really just need an implementation of ensureDestPVC() that will Get() and if not found Create(). In both cases, returning (*PVC, error).

kubectl-volsync/cmd/migration_create.go
```diff
@@ -36,70 +91,319 @@ var migrationCreateCmd = &cobra.Command{
...
+			klog.Infof("Storage class is not specified as part of cmd arg, binding to default storage class")
+			mrd.Destination.StorageClassName = nil
+		} else {
+			mrd.Destination.StorageClassName = mc.StorageClassName
+		}
+
+		mrd.Destination.DestinationPVC = &mc.PVCName
+	} else {
+		mrd.Destination.DestinationPVC = &mc.PVC.Name
+		mrd.Cluster = mc.PVC.ClusterName
```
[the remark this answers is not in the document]
v-harihar: okay. Line 233 is redundant or not required. Removing it

kubectl-volsync/cmd/migration_create.go
v-harihar: sure

kubectl-volsync/cmd/migration_create.go
v-harihar: okay. Keeping only DestinationPVC and service type assignment and removing irrelevant assignments.

kubectl-volsync/cmd/migration_create.go
v-harihar: The function `getDestinationPVC` is called at the start of `Run`. If PVC is not found and capacity is not provided, we ask the user to provide `capacity`. 
Then in the later part of `Run`, we do `ensureDestPVC()`. If we do `getDestinationPVC` as part of `ensureDestPVC` then we need to do some cleanup like delete ns etc as we come to `ensureDestPVC` after `ensureNamespace`.

kubectl-volsync/cmd/migration_create.go
JohnStrunk: Once the PVC exists, no I don't think we need the values any more.

My objection isn't really about saving vs not saving them. It's about where they are being saved. They are being put in a `volsyncv1alpha1.ReplicationDestinationRsyncSpec` which implies to the reader (though I realize the code doesn't) that they will be used when creating the ReplicationDestination object.

I'll leave it up to you about whether to save the values... just not in that struct because it makes the code hard to understand.

kubectl-volsync/cmd/migration_create.go
JohnStrunk: That's a good point. It is nice to not have to clean up the NS creation.

kubectl-volsync/cmd/migration_create.go
```diff
@@ -338,32 +330,36 @@ func (mc *migrationCreate) createDestinationPVC(ctx context.Context) (*corev1.Pe
...
+func (mc *migrationCreate) getDestinationPVC(ctx context.Context) (*corev1.PersistentVolumeClaim, error) {
 	destPVC := &corev1.PersistentVolumeClaim{}
 	pvcInfo := types.NamespacedName{
 		Namespace: mc.Namespace,
 		Name:      mc.PVCName,
 	}
 	err := mc.clientObject.Get(ctx, pvcInfo, destPVC)
 	if err != nil {
-		return nil
+		if strings.Contains(err.Error(), "not found") {
```
JohnStrunk, line 341: See: `client.IgnoreNotFound(err)`

kubectl-volsync/cmd/migration_create.go
v-harihar: > Once the PVC exists, no I don't think we need the values any more.


Makes sense

kubectl-volsync/cmd/migration_create.go
v-harihar: done

kubectl-volsync/cmd/migration_create_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+	It("Verify namespace creation with no pre-existing namespace", func() {
+		mrd.Namespace = mc.Namespace
+		ns, err := mc.ensureNamespace(cmd.Context())
+		Expect(err).ToNot(HaveOccurred())
+		Expect(ns).NotTo(BeNil())
+
+	})
+
+	It("Verify PVC creation with/without pre-existing PVC", func() {
```
JohnStrunk, line 103: I'm having trouble figuring out what portion of the code is being tested here. There seems to be a good bit of duplication from `Run()`. Experiment: Can the code or test be restructured to make it more clear?

kubectl-volsync/cmd/migration_create_test.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	return nil
+}
+
+func (mc *migrationCreate) deleteNamespace(ctx context.Context, ns *corev1.Namespace) error {
+	if err := mc.clientObject.Delete(ctx, ns); err != nil {
+		return err
+	}
+	err := wait.PollImmediate(5*time.Second, 2*time.Minute, func() (bool, error) {
```
JohnStrunk, line 191: You don't need to wait for completion of `Delete()`, and it may never complete due to the test API server.

Deletions can just be written as:
```go
Expect(k8sClient.Delete(ctx, obj)).To(Succeed())
```

JohnStrunk (commented):
Moving on to reviewing the tests...

It's also time to make sure CI is green. `make test` should pass.

kubectl-volsync/cmd/migration_create_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+		err = mc.parseCLI(cmd)
+		Expect(err).NotTo(HaveOccurred())
+
+		k8sClient, err = newClient("")
+		Expect(err).ToNot(HaveOccurred())
+		Expect(k8sClient).ToNot(BeNil())
+		mc.clientObject = k8sClient
+
+	})
```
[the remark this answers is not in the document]
v-harihar: sure

kubectl-volsync/cmd/migration_create_test.go
```diff
@@ -0,0 +1,281 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		err = mc.parseCLI(cmd)
+		Expect(err).To(HaveOccurred())
+	})
+
+	It("Verify namespace creation with no pre-existing namespace", func() {
+		mrd.Namespace = mc.Namespace
+		ns, err := mc.ensureNamespace(cmd.Context())
+		Expect(err).ToNot(HaveOccurred())
+		Expect(ns).NotTo(BeNil())
```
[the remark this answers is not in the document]
v-harihar: sure

kubectl-volsync/cmd/migration_create_test.go
v-harihar: Okay. I think I need to test functions than functionality. Am I right?

JohnStrunk: Correct.
For full functional testing, we will add cli-based tests to the e2e suite.

kubectl-volsync/cmd/migration_create_test.go
v-harihar: But the subsequent tests will fail if the deletion status is still "Terminating" when a new test starts.

For ex:
```

I0123 15:49:21.674848  131341 migration_create.go:263] Namespace: "dest" is found, proceeding with the same
• Failure [0.228 seconds]
migration
/home/vharihar/project/volsync/kubectl-volsync/cmd/migration_create_test.go:20
  Ensure replicationdestination creation [It]
  /home/vharihar/project/volsync/kubectl-volsync/cmd/migration_create_test.go:138

  Unexpected error:
      <*errors.StatusError | 0xc0002e0140>: {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "persistentvolumeclaims \"volsync\" is forbidden: unable to create new content in namespace dest because it is being terminated",
              Reason: "Forbidden",
              Details: {
                  Name: "volsync",
                  Group: "",
                  Kind: "persistentvolumeclaims",
                  UID: "",
                  Causes: [
                      {
                          Type: "NamespaceTerminating",
                          Message: "namespace dest is being terminated",
                          Field: "metadata.namespace",
                      },
                  ],
                  RetryAfterSeconds: 0,
              },
              Code: 403,
          },
      }
      persistentvolumeclaims "volsync" is forbidden: unable to create new content in namespace dest because it is being terminated
  occurred
```
 Is it not a good idea to wait till completion and start a fresh test?

kubectl-volsync/cmd/migration_create_test.go
```diff
@@ -0,0 +1,281 @@
...
+		Expect(mc.PVC).ToNot(BeNil())
+
+		// PVC creation should fail if PVC already exists
+		mc.PVC, err = mc.ensureDestPVC(cmd.Context())
+		Expect(err).NotTo(HaveOccurred())
+		Expect(mc.PVC).NotTo(BeNil())
+
+	})
+
+	It("Verify migration destination creation with/without pre-existing migration destination", func() {
```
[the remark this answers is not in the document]
v-harihar: done

kubectl-volsync/cmd/migration_create_test.go
v-harihar: sure

kubectl-volsync/cmd/migration_create_test.go
```diff
@@ -0,0 +1,281 @@
...
+		k8sClient, err = newClient("")
+		Expect(err).ToNot(HaveOccurred())
+		Expect(k8sClient).ToNot(BeNil())
+		mc.clientObject = k8sClient
+
+	})
+
+	AfterEach(func() {
+		if ns.Name != "" {
+			Expect(mc.deleteNamespace(cmd.Context(), ns)).To(Succeed())
```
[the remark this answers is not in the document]
v-harihar: sure

kubectl-volsync/cmd/migration_create_test.go
```diff
@@ -0,0 +1,281 @@
...
+			init = true
+		}
+
+		err := migrationCmdArgsSet(cmd, migrationCmdArgs)
+		Expect(err).ToNot(HaveOccurred())
+
+		err = mc.parseCLI(cmd)
+		Expect(err).NotTo(HaveOccurred())
+
+		k8sClient, err = newClient("")
```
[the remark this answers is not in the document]
v-harihar: sure

v-harihar:
> Moving on to reviewing the tests...
> 
> It's also time to make sure CI is green. `make test` should pass.

I am not sure the coverage file needs to be explicitly created. Even on local run it fails with similar error

```
KUBEBUILDER_ASSETS="/home/runner/.local/share/kubebuilder-envtest/k8s/1.22.1-linux-amd64" /home/runner/work/volsync/volsync/bin/ginkgo -progress -randomizeAllSpecs -randomizeSuites -slowSpecThreshold 30 -p -cover -coverprofile cover.out -outputdir . ./...
[1642935250] Rsync mover - 38 specs - 2 nodes •••••••••••••••••••••••••••••••••••••• SUCCESS! 29.357647612s 
[1642935250] Controller - 89 specs - 2 nodes ••••••••••••••••••••P•••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••• SUCCESS! 1m29.907525823s 
[1642935250] Volumehandler - 8 specs - 2 nodes •••••••• SUCCESS! 12.83253519s 
[1642935250] Restic mover - 33 specs - 2 nodes ••••••••••••••••••••••••••••••••• SUCCESS! 15.776819564s 
[1642935250] kubectl-volsync Suite - 20 specs - 2 nodes ••••
	 -------------------------------------------------------------------
	|                                                                   |
	|  Ginkgo timed out waiting for all parallel nodes to report back!  |
	|                                                                   |
	 -------------------------------------------------------------------

 cmd timed out. path: ./kubectl-volsync/cmd
path is /home/runner/work/volsync/volsync
Unable to read coverage file  to combine, open : no such file or directory

Ginkgo ran 5 suites in 3m12.162999893s
Test Suite Failed
```

JohnStrunk:
> > Moving on to reviewing the tests...
> > It's also time to make sure CI is green. `make test` should pass.
> 
> I am not sure the coverage file needs to be explicitly created. Even on local run it fails with similar error

The timeout error is because the tests are broken (in a way that is more serious than just an `Expect` failing)--- and because of the error, coverage files aren't properly generated.

v-harihar:
> > > Moving on to reviewing the tests...
> > > It's also time to make sure CI is green. `make test` should pass.
> > 
> > 
> > I am not sure the coverage file needs to be explicitly created. Even on local run it fails with similar error
> 
> The timeout error is because the tests are broken (in a way that is more serious than just an `Expect` failing)--- and because of the error, coverage files aren't properly generated.

Is it possible to increase the timeout value to some bigger value?

kubectl-volsync/cmd/migration_create.go
```diff
@@ -36,70 +79,321 @@ var migrationCreateCmd = &cobra.Command{
...
+		Namespace: mrd.Namespace,
+		Name:      mrd.RDName,
+	}
+	rd := &volsyncv1alpha1.ReplicationDestination{}
+	err := wait.PollImmediate(5*time.Second, 2*time.Minute, func() (bool, error) {
+		err := mc.clientObject.Get(ctx, nsName, rd)
+		if err != nil {
+			return false, err
+		}
+		if rd.Status == nil {
```
JohnStrunk, line 378: Rsync field is a pointer, so need a `nil` check

```suggestion
		if rd.Status == nil || rd.Status.Rsync == nil {
```

JohnStrunk (commented):
Other than the nil check, I think this is ready.
Please squash and force push after fixing the nil check.

PR #116, opened 2021-12-08
Add (de-)serialization for resource.Quantity

JohnStrunk:
resource.Quantity doesn't serialize/deserialize properly via Viper because its fields are private. This adds SetData and GetData calls that:

1) Flatten the structure so that the quantity's string value can be set without destroying the rest of the data
2) Add a decoding hook to properly unmarshal the Capacity from a string

**Related issues:**

### DIFF
--- a/kubectl-volsync/cmd/relationship.go
+++ b/kubectl-volsync/cmd/relationship.go
@@ -21,10 +21,13 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"reflect"
 
 	"github.com/google/uuid"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"gopkg.in/yaml.v2"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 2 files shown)


v-harihar (approved):

v-harihar (commented):
@JohnStrunk I tried to verify this in my patch and the quantity does not get written in the right format.

```
func (mr *migrationRelationship) Save() error {
       err := mr.SetData(mr.data)
       if err != nil {
               return err
       }
        return mr.Relationship.Save()
 }
```
config file contents
```
# cat /root/.volsync/v3.yaml 
data:
  destination:
    cluster: ""
    destination:
      address: 10.96.141.201
      replicationdestinationvolumeoptions:
        accessmodes:
        - ReadWriteOnce
        capacity:
          format: BinarySI
        copymethod: Direct
        storageclassname: csi-hostpath-sc
      servicetype: ClusterIP
      sshkeys: volsync-rsync-dest-src-dest-volsync-migration-dest
    mdname: dest-volsync-migration-dest
    namespace: dest
    pvcname: volsync
    sshkeyname: ""
  version: 1
id: 9977280a-4b22-4367-a722-c0da4f7d6b92
type: migration
```

Please correct me if wrong

PR #122, opened 2022-01-06
CLI: create config directory if it doesn't exist

JohnStrunk:
Creates the configuration directory (default: `~/.volsync`) if it doesn't already exist. This fixes the problem where saving the relationship file would fail if the directory wasn't there.

**Related issues:**


aspandey (approved):

v-harihar (approved):
lgtm

PR #123, opened 2022-01-10
CLI replication create & delete

JohnStrunk:
Adds the `create` and `delete` commands to the `kubectl-volsync replication` sub-command

**Related issues:**


v-harihar:
@JohnStrunk , This PR provides the skeleton for replication CLI and not the complete implemenation. Am I right ?

JohnStrunk:
Just create and delete. There are 2 others: set source/destination, and one that has schedule/sync. It's the last of those that actually apply objects to the cluster.

v-harihar:
> Just create and delete. There are 2 others: set source/destination, and one that has schedule/sync. It's the last of those that actually apply objects to the cluster.

Makes sense.

v-harihar (approved):
lgtm

v-harihar:
There are few minor comments like moving the imports and struct member names. But those can be taken care/corrected in full implementation of replication cmds.

PR #124, opened 2022-01-12
Adding usage example for rsync in a cross cluster scenario

tesshuflower:

Usage docs - additional rsync example for cross cluster scenario.  We think cross cluster is most likely the 
most common user scenario at this point.

**Related issues:**


tesshuflower:
@JohnStrunk would you mind taking a look at these doc updates?  I believe I added the link properly in the index, but I wasn't sure exactly how to confirm.

PR #139, opened 2022-02-09
Use viper to allow both cmd line and env vars for mover container imgs

tesshuflower:

In order to complete CSV packaging as an add-on, we want to be able to override mover image locations via Env vars.

Currently mover images can be specified as command line parameters.  This PR uses viper to allow the current command line parameters to work, but if they are not set, then use the corresponding environment variable.  If neither are set, the default will be used.


controllers/mover/rclone/builder.go
```diff
@@ -33,21 +34,38 @@ import (
...
 
 var _ mover.Builder = &Builder{}
 
-func Register() {
-	flag.StringVar(&rcloneContainerImage, "rclone-container-image",
-		defaultRcloneContainerImage, "The container image for the rclone data mover")
+func init() {
+	// Set default rclone container image - will be used if both command line flag and env var are not set
+	viper.SetDefault(rcloneContainerImageFlag, defaultRcloneContainerImage)
+}
```
JohnStrunk, line 49: Can this just be done in the Register() function instead of having an init?

main.go
```diff
@@ -23,6 +23,9 @@ import (
 	"os"
 	"runtime"
 
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+
```
JohnStrunk, line 28: Please put these in the "snapv1" block w/ the other external packages

controllers/mover/rclone/builder.go
tesshuflower: @JohnStrunk I've made another commit to re-arrange a bit (including the change you suggested above) and allow unit tests to set their own viper & flagset to hopefully avoid issues with parallel tests.  Please let me know what you think, it does end up a little more convoluted, but hopefully helps with tests.  It still may be possible to get errors with the env var tests, although I haven't encountered any yet.

JohnStrunk: What about something like having `RegisterWithFlags()`?
so we get:
```golang
func Register() error {
    return RegisterWithFlags(flag.CommandLine, viper.GetViper())
}

// This could even be little "registerWithFlags" since it's only being accessed by tests
func RegisterWithFlags(flags, viper) error {
    b := &Builder {
        viper: viper,
        flags: flags
    }
    // initFlags contents
    mover.Register(b)
}
```

I think this could get rid of the getViper/getFlagset functions and just let everything directly reference b.viper, etc. For example:
```golang
func (rb *Builder) getRcloneContainerImage() string {
	return rb.viper.GetString(rcloneContainerImageFlag)
}
```

Does this seem less convoluted?

tesshuflower: Hey John, thanks for the suggestion, I do like this idea, but one question:  One thing I was trying to avoid was calling register multiple times, resulting in the overall mover.Catalog getting the same builder appended to it multiple times by different tests.  Generally most of the mover tests don't need register (we have one specific Register() test which is good), but I ended up initing the flags again for those tests to force the mover image to be initialized.

What do you think about me updating initFlags (can rename to newBuilder or something of the sort) to work like your registerWithFlags() suggestion but just without the mover.Register() call?

i.e. 
```
func Register() error {
  b := newBuilder(flag.CommandLine, viper.GetViper())
  mover.Register(b)
}

func newBuilder(flags, viper) error {
      b := &Builder {
        viper: viper,
        flags: flags
    }
   b.viper.SetDefault(rsyncContainerImageFlag, defaultRsyncContainerImage)
   ... etc
}
```

JohnStrunk: That works. :+1:

tesshuflower: Thanks!  I've updated the PR again with this suggestion

Issue #140, opened 2022-02-10
Add a flag that will activate OpenShift SCC when autodetection isn't possible


larsks:
**Describe the feature you'd like to have.**

When installing the Helm chart using ArgoCD, Helm isn't able to query the remote cluster to resolve any conditionals in the chart, which means that when deploying into our OpenShift environment the necessary SCC was not created. Let's add a chart parameter that can be set explicitly to trigger generation of the SCC.

**What is the value to the end user? (why is it a priority?)**

This would allow the end user to install volsync in situations in which it wasn't possible to query the target cluster.

PR #146, opened 2022-02-24
Add option to force deployment of mover SCC

JohnStrunk:
Adds a configuration option to the Helm chart that forces deployment of the custom mover SecurityContextConstraint. Normally, the existing autodetection is sufficient, but in cases where users want to template the chart without querying the cluster, it may be necessary to force the generation of the SCC.

**Related issues:**
Fixes: #140

### DIFF
--- a/helm/volsync/README.md
+++ b/helm/volsync/README.md
@@ -140,3 +140,7 @@ on the command line or via a custom `values.yaml` file.
   - Allows applying tolerations to the operator pod
 - `affinity`: none
   - Allows setting the operator pod's affinity
+- `openshift.forceSCC`: false
+  - Forces the VolSync mover SecurityContextConstraint to be deployed even if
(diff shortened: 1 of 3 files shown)


JohnStrunk:
@larsks Does this match what you're looking for?

```console
# Autodetect does not provide the SCC when templating (existing behavior)
$ helm template -n foo volsync ./helm/volsync | grep SecurityContextConstraints
$

# Supplying force option generates the SCC
$ helm template -n foo --set openshift.forceSCC=true volsync ./helm/volsync | grep SecurityContextConstraints
kind: SecurityContextConstraints
```

JohnStrunk:
/cc @tesshuflower 

I couldn't think of a better place to put the config flag to keep the chart options organized. If you have a better suggestion, let me know-- otherwise, this is ready.

tesshuflower (approved):
/lgtm

PR #147, opened 2022-02-28
Upgrade operator-sdk v1.15 -> v1.18

JohnStrunk:
Brings the operator-sdk scaffolding up to v1.18

**Related issues:**


JohnStrunk:
/hold

Latest operator-sdk requires golang 1.17, but we're building w/ the latest ubi8/go-toolset which is 1.16. We need a solution that is go 1.17, but also fips compatible.

JohnStrunk:
/close
Will try this again in a different PR

PR #156, opened 2022-03-02
Commit missed generated bundle scorecard config files

tesshuflower:

config.yaml files were in .gitignore, so some generated files from initial bundle generation got missed.
This means subsequent make bundle commands could fail as they are missing some expected files.

I modified .gitignore to only ignore /config.yaml  - Is this correct?

**Related issues:**


JohnStrunk (commented):
/lgtm

PR #166, opened 2022-03-16
Operator namespace

tesshuflower:
Removed suggested namespace in CSV of `volsync-system`.  This means OLM will only allow the install of this operator into the `openshift-operators` namespace.

Also inserted the generic "Red Hat" icon into the CSV.  This can be updated later if we get our own custom icon.


JohnStrunk:
/lgtm

Docs failure isn't due to this PR

PR #157, opened 2022-03-02
Document migration CLI sub-command

JohnStrunk:
**Describe what this PR does**
- [x] Document `kubectl-volsync migration`
- [x] Fix references to external migration
- [x] Remove old external sync script

**Is there anything that requires special attention?**

**Related issues:**
Fixes: #154 
Depends on: #141 (docs)

PR #178, opened 2022-03-21
Increase operator responsiveness

JohnStrunk:
Reconcile operations w/ a long latency cause the controller to be less responsive to other CRs in the system. This PR:
- Increases the CPU limit to allow CPU-intensive operations (key generation) to complete quicker
- Increases the max number of simultaneous reconciles from 1 for each CR type to 100 per type to enable cross-CR parallelism

I *think* this is the cause of the intermittent CI failures of the kuttl tests in #157. The CLI tests add additional tests that cause ssh keys to be generated.

I would like to just remove the CPU limit, but that increases the risk of collateral damage if VolSync were to get into a loop.

**Related issues:**

### DIFF
--- a/bundle/manifests/volsync.clusterserviceversion.yaml
+++ b/bundle/manifests/volsync.clusterserviceversion.yaml
@@ -310,7 +310,7 @@ spec:
                   periodSeconds: 10
                 resources:
                   limits:
-                    cpu: 200m
+                    cpu: 1000m
                     memory: 300Mi
                   requests:
                     cpu: 100m
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -48,7 +48,7 @@ spec:
           periodSeconds: 10
         resources:
           limits:
-            cpu: 200m
+            cpu: 1000m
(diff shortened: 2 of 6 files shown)


JohnStrunk:
/cc @tesshuflower 
/assign @tesshuflower 

@tesshuflower What do you think of adding this to 0.4?
The CPU limit change is benign, but I have less experience w/ the max reconciles. @guymguym, @RobotSail, and I did look through the library code a few weeks ago, and it does seem safe (no concurrent reconcile of the same CR, just parallelism across CR instances).

tesshuflower (approved):
/lgtm

tesshuflower:
> /cc @tesshuflower /assign @tesshuflower
> 
> @tesshuflower What do you think of adding this to 0.4? The CPU limit change is benign, but I have less experience w/ the max reconciles. @guymguym, @RobotSail, and I did look through the library code a few weeks ago, and it does seem safe (no concurrent reconcile of the same CR, just parallelism across CR instances).

My vote is to add this to 0.4!

PR #185, opened 2022-03-23
Switch to snapshot.storage.k8s.io/v1

### LINKED ISSUE #182: Switch to snapshot.storage.k8s.io/v1
Change the controller code to use the v1 version of VolumeSnapshot instead of v1beta1.

It appears that the v1beta1 API is going to be removed in kube 1.24, meaning that VolSync will no longer work for 1.24+ without this change.

**Additional context**
Today, we use v1beta1 which means that VolSync is usable on Kubernetes versions 1.17 - 1.23 (although we only test back to 1.19).
When we make this change, we will require a minimum Kubernetes version of 1.20 (when v1 became available). While upstream Kube no longer supports the older versions, OCP, AKS, and GKE still support the equivalent of 1.19 until later this year.

JohnStrunk:
This moves VolSync from snapshot.storage.k8s.io/v1beta1 to snapshot.storage.k8s.io/v1

This will increase the minimum Kubernetes version required for VolSync to Kubernetes 1.20

**Related issues:**
Fixes #182


JohnStrunk:
/cc @tesshuflower 
/assign @tesshuflower 

I think this is ready for review. I think I updated all the places where we reference kube versions. Let me know if you find any I missed.

tesshuflower (approved):
/lgtm

tesshuflower:
@JohnStrunk Not critical, but should we cleanup some of the code in the helper scripts in hack now that we only support kube >= 1.20?  This sort of stuff:  https://github.com/backube/volsync/blob/main/hack/setup-kind-cluster.sh#L100-L114

JohnStrunk:
^^^ Will try to CP again after #188

Issue #67, opened 2021-08-30
Better feedback on problems


JohnStrunk:
**Describe the feature you'd like to have.**
It would be good if the operator provided more actionable feedback in the CR status so that it was easier to troubleshoot when things aren't working as expected.

**What is the value to the end user? (why is it a priority?)**
End users would be able to more quickly diagnose what's not working and why so that they can get their environment up and running more quickly.

**How will we know we have a good solution? (acceptance criteria)**
- The `.status.conditions` section of the Source and Destination CRs should provide information about where in the reconcile sequence the operator is.
- *Anything else?*

**Additional context**
In #66, the operator wasn't able to complete the sync iteration because the snapshot couldn't be created. Unfortunately, there was no error (or status information) generated to indicate that this was the problem. Do the the async nature of kube, it's going to be hard to tell whether something is broken/misconfigured or just taking a while, but by having the operator expose more status information, it can at least narrow down where to look when things aren't progressing as quickly as expected.

Brainstorming a bit:  
We have the "Synchronizing" condition:
```golang
const (
	ConditionSynchronizing     string = "Synchronizing"
	SynchronizingReasonSync    string = "SyncInProgress"
	SynchronizingReasonSched   string = "WaitingForSchedule"
	SynchronizingReasonManual  string = "WaitingForManual"
	SynchronizingReasonCleanup string = "CleaningUp"
)
```
While the expressiveness of "reason" is fairly limited (and I don't know that I want to expand the number of reason codes), the free form messages could be more descriptive:
- "Creating keys"
- "Creating Snapshot"
- "Waiting for Job to complete"
- etc.

JohnStrunk:
This was discussed in this week's meeting. Some thoughts/ideas:
* @shawn-hurley mentioned the possibility of having a webhook to validate and reject w/ a message about the misconfiguration
* @alaypatel07 suggested posting events
  * This would be very similar to the Pod's model where you get events when things are taking a long time (e.g., volume mount). Like Pod events, we may not be able to say whether it's broken or just taking a long time, but it would expose what's causing the delay.
* @alaypatel07 also suggested adding more Conditions to the RS/RD objects for detectable misconfigurations (like multiple default VSClasses)

v-harihar:
@JohnStrunk I am taking look at webhooks and its applications. Please provide any inputs if you have

JohnStrunk:
/close
We have been making progress on this in a number of other issues/PRs

PR #194, opened 2022-04-08
Give movers the ability to send Events

JohnStrunk:
This expands the Builder/Mover interface to give Mover implementations access to an events.EventRecorder object so that they can publish Events.

controller-runtime is still using the old corev1.Event interface (record.EventRecorder), but this set of changes exposes the new events.EventRecorder for events.k8s.io/v1 Event objects. This PR contains an adapter object that allows us to use the new interface while publishing through the interface that controller-runtime gives us. Once controller-runtime provides the new interface, we can remove our adapter.
The adapter is a copy of the one in `k8s.io/client-go/tools/events`. I am duplicating it so that we can find a way to include the "related" and "action" fields that the default adapter drops.

**Related issues:**
Related to #67


JohnStrunk:
Example of publishing Events (from `kubectl describe replicationdestination/...`)
```
Events:
  Type     Reason        Age    From                Message
  ----     ------        ----   ----                -------
  Normal   MoverStarted  6m51s  volsync-controller  starting rsync mover related=Job/volsync-rsync-dst-replicationdestination-sample action=CreatingMoverJob
```

JohnStrunk:
/cc @tesshuflower 

This is ready for a review.
It's just the scaffolding--- #202 shows how it will be used.

controllers/replicationsource_controller.go
```diff
@@ -139,7 +143,9 @@ func reconcileSrcUsingCatalog(
 	// Search the Mover catalog for a suitable data mover
 	var dataMover mover.Mover
 	for _, builder := range mover.Catalog {
-		if candidate, err := builder.FromSource(sr.Client, logger, instance); err == nil && candidate != nil {
+		candidate, err := builder.FromSource(sr.Client, logger,
+			record.NewEventRecorderAdapter(sr.EventRecorder), instance)
```
tesshuflower, line 147: @JohnStrunk Just for my knowledge, what is the purpose of wrapping with NewEventRecorderAdapter()?

tesshuflower (approved):
/lgtm

controllers/replicationsource_controller.go
JohnStrunk: This is the "new" vs "old" Event interface. The new interface is [`events.EventRecorder`](https://pkg.go.dev/k8s.io/client-go@v0.23.5/tools/events#EventRecorder), and the old one is [`record.EventRecorder`](https://pkg.go.dev/k8s.io/client-go@v0.23.5/tools/record#EventRecorder). The new interface publishes events.k8s.io/Event objects whereas the old one is v1/Event. The k8s.io version adds a couple of additional fields for "related" object and "action". Both event types look the same when doing a `kubectl describe <object>`.

The adapter allows us to use the new interface and provide those additional fields even though controller-runtime is still using the old interface. Those extra fields will be dropped by the adapter, though we could provide our own adapter to tack them onto the message if we wanted. I had written a custom adapter that added `related` & `action` as k/v pairs to the `message` field, but I decided to drop it from this PR since it wasn't really providing any useful information.

Once controller-runtime includes the new interface, we'll drop the adapter.

PR #203, opened 2022-04-19
Update to golang 1.18

JohnStrunk:
Updates the version of golang to 1.18

We have been stuck on golang 1.16 due to `ubi8/go-toolset` not supporting anything more recent. Given that the kube ecosystem is now requiring 1.18 and there doesn't seem to be any movement in go-toolset, this PR moves us back to the official golang builder images. The unfortunate part is that we can't build w/ the hooks for goboring to make sure we're still ok w/ fips in the product builds.
I see no reasonable way to get both. :man_shrugging: 

There's also a type assertion fix in here, which is really a fix for my earlier "fix" that I believe made the test non-useful.

golangci-lint had to be upgraded to get support for go 1.18.

**Related issues:**
Should unblock #147


JohnStrunk:
/hold

Looks like we'll need a newer version of Syncthing: 1.20.0 or 1.19.3 to pick up syncthing/syncthing#8231

```
/go/pkg/mod/github.com/lucas-clemente/quic-go@v0.23.0/internal/qtls/go118.go:5:13: cannot use "quic-go doesn't build on Go 1.18 yet." (untyped string constant) as int value in variable declaration
```

RobotSail:
Fans of Generics rejoice!

tesshuflower (approved):
/lgtm

Issue #201, opened 2022-04-18
CLI replication: set-* should completely overwrite old settings


JohnStrunk:
**Describe the bug**
It is possible to modify the source or destination settings by using the `set-source` or `set-destination` commands followed by a call to `schedule` or `sync`. However, the `set-*` doesn't completely erase the old settings, making it impossible to remove optional parameters.

**Steps to reproduce**
```console
$ ./bin/kubectl-volsync replication -r test create
$ ./bin/kubectl-volsync replication -r test set-source --pvcname src/datavol
$ ./bin/kubectl-volsync replication -r test set-destination --destination dst/datavol --volumesnapshotclass bad
$ ./bin/kubectl-volsync replication -r test schedule --cronspec '*/4 * * * *'

# VSC of the dest is wrong... erase it to have the default
$ ./bin/kubectl-volsync replication -r test set-destination --destination dst/datavol

# From `kubectl describe`:
Spec:
  Rsync:
    Access Modes:
      ReadWriteOnce
    Capacity:                    3Gi
    Copy Method:                 Snapshot
    Service Type:                ClusterIP
    Volume Snapshot Class Name:  bad                      <=== still there
```

**Expected behavior**

**Actual results**

**Additional context**
Current workaround is to `delete` the relationship and recreate it.

PR #207, opened 2022-04-26
CLI: Properly overwrite deleted config options

JohnStrunk:
When updating the CLI relationship files, it was not possible to delete an option once it had been written into the file.
Example:
```console
$ ./bin/kubectl-volsync replication -r test create
$ ./bin/kubectl-volsync replication -r test set-destination --destination dst/datavol --volumesnapshotclass bad

# VSC of the dest is wrong... erase it to have the default
$ ./bin/kubectl-volsync replication -r test set-destination --destination dst/datavol
```

The above set of commands should have removed the VSC from the relationship file, but it did not. This is due to Viper not supporting the removal of keys. When we attempt to set a key to `nil`, it just retains the old value.

With this patch, we actually create a new Viper instance and swap it out in order to get the intended effect.

**Related issues:**
Fixes: #201

### DIFF
--- a/kubectl-volsync/cmd/relationship.go
+++ b/kubectl-volsync/cmd/relationship.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path"
 	"reflect"
+	"strings"
 
 	"github.com/google/uuid"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


v-harihar (approved):
/lgtm

PR #210, opened 2022-05-02
Factor out the state machine

### LINKED ISSUE #200: Condition Synchronizing only added after 1st sync
**Describe the bug**
The Condition "Synchronizing" is only added to the `.status` after the 1st sync finishes. This makes it hard to tell what's going on during the initial sync iteration.

**Steps to reproduce**
Create an RS/RD and `kubectl get ...` it.

**Expected behavior**

**Actual results**
```yaml
  status:
    conditions:
    # No "type: Synchronizing" in the list
    - lastTransitionTime: "2022-04-18T15:56:54Z"
      message: Reconcile complete
      reason: ReconcileComplete
      status: "True"
      type: Reconciled
    lastSyncStartTime: "2022-04-18T15:56:52Z"
    rsync:
      address: 10.96.28.227
      sshKeys: volsync-rsync-dst-src-datavol
```

**Additional context**
I haven't looked yet, but I suspect this is due to only updating the condition on phase change (sync vs cleanup). We just need to add a check in there to add it if it's not present.

JohnStrunk:
This factors out the Sync  :left_right_arrow: Cleanup state machine from the RS & RD code.

There was a lot of duplication between RS & RD, plus the logic was a bit convoluted for determining which state we should be in and how to update the various status indicators.

This also removes the "Reconciled" status condition that has led to user confusion. Instead, errors will be reported via a new reason code on the the Synchronizing condition (and a status of `false`).

**Related issues:**
I came to the conclusion that the overhaul was needed while trying to fix #200. While it was easy to figure out why the problem was happening, based on the associated comments, it looks like the logic was intentional.
Fixes #200


JohnStrunk:
/cc @tesshuflower

This is ready for review.

The replication*_controller.go diffs look terrible because gh tries to match up the changes. The files are much easier to look at in isolation.

controllers/replicationdestination_controller.go
```diff
@@ -232,153 +170,124 @@ func (r *ReplicationDestinationReconciler) SetupWithManager(mgr ctrl.Manager) er
...
-		metrics.OutOfSync.Set(1)
+	metrics := newVolSyncMetrics(prometheus.Labels{
+		"obj_name":      rd.Name,
+		"obj_namespace": rd.Namespace,
+		"role":          "destination",
+		"method":        dataMover.Name(),
+	})
+
+	if rd.Status == nil {
+		rd.Status = &volsyncv1alpha1.ReplicationDestinationStatus{}
```
tesshuflower, line 188: The status is still initialized in the Reconcile() - should we remove one of them? (same for replicationsource_controller)

controllers/replicationdestination_controller.go
```diff
@@ -93,36 +103,40 @@ func (r *ReplicationDestinationReconciler) Reconcile(ctx context.Context, req ct
...
-	}
+	rdm, err := newRDMachine(inst, r.Client, logger, record.NewEventRecorderAdapter(r.EventRecorder))
 
-	// Set reconcile status condition
-	if err == nil {
+	// Using only external method
+	if errors.Is(err, mover.ErrNoMoverFound) && inst.Spec.External != nil {
+		return ctrl.Result{}, nil
+	}
+	// Both internal and external methods defined
```
tesshuflower, line 112: These checks for external methods, movers etc - could they be moved into the newRDMachine or some sort of sm.Validate() call?  I think the checks seem common to replicationsource & dest with multiple methods/external etc.

controllers/replicationdestination_controller.go
```diff
@@ -93,36 +103,40 @@ func (r *ReplicationDestinationReconciler) Reconcile(ctx context.Context, req ct
...
+			Reason:  volsyncv1alpha1.SynchronizingReasonError,
 			Message: err.Error(),
 		})
 	}
 
+	// All good, so run the state machine
+	if err == nil {
+		result, err = sm.Run(ctx, rdm, logger)
+		// Retrieve any Status info modified by the state machine
+		inst.Status = rdm.ReplicationDestination.Status
```
tesshuflower, line 137: Is there a reason we need a copy of the ReplicationDestination in the rdMachine?  Then we wouldn't need this extra step to always copy over the status at the end

controllers/replicationdestination_controller.go
```diff
@@ -93,36 +103,40 @@ func (r *ReplicationDestinationReconciler) Reconcile(ctx context.Context, req ct
...
+	}
+
```
tesshuflower, line 139: I can't put the comment in the correct place here - but are line 145-152 needed anymore? 

```
if !inst.Status.NextSyncTime.IsZero() {
		// ensure we get re-reconciled no later than the next scheduled sync
		// time
		delta := time.Until(inst.Status.NextSyncTime.Time)
		if delta > 0 {
			result.RequeueAfter = delta
		}
	}
```

I think this may already be figured out when running sm.Run() - we should be able to rely on the result at this point to have requeueAfter properly set?

controllers/statemachine/machine.go
```diff
@@ -0,0 +1,277 @@
...
+			return ctrl.Result{}, err
+		}
+		if missed {
+			r.SetOutOfSync(true)
+		}
+	}
+
+	var result ctrl.Result
+	var err error
+	switch currentState(r) {
```
tesshuflower, line 66: I might be missing something here - but do we have a potential to start synchronizing out of schedule here?

I think previously there was always a check to see if we should sync (similar to the new `shouldSync()` func) prior to synchronizing, but I don't see it anymore.

If the reconcile is triggered by a change to some resource the RD/RS owns, or even something like the controller restarting (and I think there was a regular sync interval added at some point too just in case), then it seems we could enter into synchronizing (by running `doSynchronizingState()`) even when technically we're in "idle" state (for example when the next sync time hasn't hit yet).  Edit:  I suppose this could happen with trigger as well - if the trigger hasn't changed, we could possibly run a second sync with the same trigger?

tesshuflower (commented):
I don't think I've really gone into the detail of the schedule stuff yet as it's quite complicated, but did put some comments/concerns.

controllers/replicationdestination_controller.go
JohnStrunk: Maybe not... Let me take a look at switching to a *RD instead of embedding it. I embedded it originally so that the methods would be exposed, but they may not need to be.

controllers/replicationdestination_controller.go
JohnStrunk: It's probably redundant. I'll remove it.

controllers/replicationdestination_controller.go
JohnStrunk: It's done in Reconcile() so that we can set the Condition to error if necessary.
I had also added it to here because none of the methods check before they access status, but that's a good point that we're guaranteed it won't be nil due to Reconcile. Will remove this one.

controllers/statemachine/machine.go
JohnStrunk: We (are supposed to) stay locked in a state until one of the transitionToX() functions gets called.
"Idle" is really just cleanup cleanup. We stay in cleanup, but report that we're idle (L130-L144).
So, once we're in cleanup, if the mover says it's complete, then we call shouldSync(), and that's what moves us to synchronizing (L125). Extra reconciles just cause additional checks on the resources and return Complete but have no effect unless we shouldSync().

tesshuflower: aha, yes that's the part I didn't catch - after completing synchronizing the lastSyncStartTime should get cleared, so it will go to cleanup instead of synchronizing again.  Thx for the explanation

controllers/replicationdestination_controller.go
JohnStrunk: done

controllers/replicationdestination_controller.go
JohnStrunk: done

controllers/replicationdestination_controller.go
JohnStrunk: done

controllers/replicationdestination_controller.go
JohnStrunk: I agree that it's redundant across RS/RD, but I'm coming up empty on how to fix it. I could use some suggestions:

- I went pretty far down the path of creating a "null" mover that registers for the external method, but the state machine still runs and updates timestamps, so that's a non-starter.
- Moving the check into newR?Machine really just moves it and doesn't reduce the duplication, best I can tell.

The unfortunate situation is that "external" is special since it doesn't use the state machine, and the logic for matching (Catalog) also doesn't seem to be re-usable.

tesshuflower: sorry to send you down a rabbit-hole there - sometimes these things look easier from the outside than they really are...

tesshuflower (approved):
/lgtm

PR #224, opened 2022-05-18
Switch containers to ubi9

JohnStrunk:
RHEL9 is GA. This switches our base containers to use ubi9 instead of ubi8.
It also cleans up the microdnf calls across the different containers.

**Related issues:**


tesshuflower (approved):
/lgtm

PR #228, opened 2022-05-24
Unpin sphinx-toolbox

JohnStrunk:
The sphinx-toolbox version had been pinned due to 2.18.0 not working.
Now that version 3 has been released, we can remove the pin.

This also fixes a problem w/ the `setup-env.sh` script where it would fail if not run from the `docs/` directory.

**Related issues:**

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -28,5 +28,6 @@ jobs:
       - name: Build docs
         run: |
           cd docs
-          . ./setup-env.sh
+          ./setup-env.sh
+          . ../.venv/bin/activate
           make html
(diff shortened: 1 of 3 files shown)


.github/workflows/documentation.yml
```diff
@@ -28,5 +28,6 @@ jobs:
       - name: Build docs
         run: |
           cd docs
-          . ./setup-env.sh
+          ./setup-env.sh
+          . ../.venv/bin/activate
```
tesshuflower, line 32: Just out of curiosity, why was this line needed? Activate appears to get sourced in in the setup-env.sh script? I'm assuming it's some shell or sequencing thing ...

tesshuflower (approved):
/lgtm

Issue #229, opened 2022-05-25
Use SonarCloud for code coverage and vulnerability scanning


tesshuflower:
**Describe the feature you'd like to have.**
RedHat ACM is using SonarCloud for all their repos, and have suggested that we do the same.

**What is the value to the end user? (why is it a priority?)**
The main benefit to start may be the easy to view interface which shows code coverage as well as vulnerabilities and more (code smells, etc).

**How will we know we have a good solution? (acceptance criteria)**
- PRs should report sonar status
- we should be able to see code coverage differences before merging a PR
- We should be able to view volsync in sonar cloud to easily see sonar's code coverage and quality data

**Additional context**

JohnStrunk:
/assign

As a part of this, I'll also remove codecov

JohnStrunk:
It's looking like SonarCloud isn't ready for prime time.
They require access to an upload token that needs to be stored in a GH secret. However, GH secrets (for obvious reasons) aren't make available in jobs run against PRs from external forks... which is... all of our PRs.
Feature request: https://portal.productboard.com/sonarsource/1-sonarcloud/c/50-sonarcloud-analyzes-external-pull-request

codecov gets around this by not needing a token. I don't know how they manage to do it securely. :man_shrugging: 

Stolostron is getting around this by having the token handled via OpenShift CI.

JohnStrunk:
/close
I think we're using as much as we can at this point. Feel free to re-open if you think there's more we should do here.

PR #230, opened 2022-05-26
Send code quality info to SonarCloud

JohnStrunk:
- Swaps out codecov for SonarCloud

- I added a special `make test-ci` target that is equivalent to the normal `make test` except that it generates output that is in the format required for SonarCloud.
  - I didn't want to change the existing target because the new format is not very readable
  - The unfortunate consequence is that these targets now need to be kept in sync

**Related issues:**
#229


JohnStrunk:
/close
There's currently no good way to make the code coverage stuff work.

PR #232, opened 2022-05-27
Enable dependabot updates

JohnStrunk:
- Enables dependabot to keep go.mod and GH actions up-to-date
- Switches the version specifier for actions to be SHA hashes instead of tags to improve security
- Increases the fetch-depth of the git checkout for the job that does code coverage to see if that improves the consistency of the coverage reports from codecov

**Related issues:**

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+---
+version: 2
+updates:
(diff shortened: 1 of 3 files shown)


tesshuflower (approved):
/lgtm

PR #249, opened 2022-06-02
Update minio version

JohnStrunk:
- 9.0.5 is no longer in the bitnami repo
- This updates us to the latest version

We should probably establish some sort of strategy for ensuring the availability of the charts we use in CI

**Related issues:**


JohnStrunk:
/cc @tesshuflower 
Looks like the latest ubi8 images (we're still using them in OpenShift CI) are [currently broken (perl can't be installed)](https://bugzilla.redhat.com/show_bug.cgi?id=2092960).

Issue #246, opened 2022-06-01
When creating a PVC from snapshot, the incorrect size may be requested


tesshuflower:
**Describe the bug**
If a PVC has a requested storage size that is different from the actual capacity (as an example a PVC requests a 10Mi size, but the capacity actually allocated ends up being 1Gi) then when creating a PVC from snapshot, the PVC may use the requested size from the PVC instead of the size indicated in the snapshot.

https://github.com/backube/volsync/blob/main/controllers/volumehandler/volumehandler.go#L440-L452

The code will look at the restoreSize in the status of the snapshot, but if it's not filled out yet (not sure yet if this is a required field or not) then it falls back to taking the requestedSize from the source PVC.

This can result in the PVC being stuck in Pending because the requested size is smaller than the restoreSize for the Snapshot.

**Additional context**

As discussed in the scrum, ideally we would do the following to fix:

- Make sure the snapshot is in Bound state before attempting to create the PVC from the snapshot
- Use the restoreSize on the snapshot (need to check if this is a mandatory field or not)
- If restoreSize is not found, instead of falling back immediately to the requested size of the source PVC, first check the pvc status for capacity, and attempt to use that first
- If all else fails, then finally fall back to the requested size of the source PVC.

Related issue:  https://github.com/backube/volsync/issues/48

PR #251, opened 2022-06-02
Pvc storage size - clone & snapshot - use capacity if possible

tesshuflower:
For pvcFromSnapshot:
  - attempt to use the restoreSize from the snapshot to determine size to create the PVC
  - if restoreSize isn't available, attempt to use the status.capacity from the origin pvc
  - fallback to using the requested storage size from origin pvc

For clone:
  - attempt to use the status.capacity from the origin pvc
  - fallback to using the requested storage size from origin pvc

- For snapshots, we had discussed making sure the snapshot is bound at the beginning of pvcFromSnapshot() - it turns out that  EnsurePVCFromSrc()  calls ensureSnapshot first and then pvcFromSnapshot (see https://github.com/backube/volsync/blob/main/controllers/volumehandler/volumehandler.go#L73-L78)  - and in fact ensureSnapshot will only return a snapshot after it's bound.  So it turns out the bound check was already there https://github.com/backube/volsync/blob/fa4bdcfb2d88d438f39bdceb1159c5a2f30fbaa6/controllers/volumehandler/volumehandler.go#L405-L414

  The issue I found when recreating is it's possible to get a snapshot with a status like this:

  ```
  {"boundVolumeSnapshotContentName":"snapcontent-b735e05f-6352-45bf-a5ab-89a87577b64f","readyToUse":false}}
  ```
  That is, bound but no restoreSize set yet

  and then later get one with restore size set - so in this sort of situation we'll always be falling back to the PVC capacity.

  We could potentially try to wait for the snapshot readyToUse: true before trying to create a PVC?

  Unfortunately I wasn't able to find any concrete information about whether these fields are mandatory or not, none are 
  specifically shown as required in the status as far as I can tell.

- For clone, one thing I see that we could hit is where a user has their storageclass set to `WaitForFirstConsumer`.  In this case if a user creates a repilcationsource for that PVC then capacity may never be filled out, and for clone we'll still fallback to using the requested size.  I wasn't sure if this is a real concern out there or not - we could potentially check that the source pvc is in Bound state before proceeding.

  note I don't think this is an issue with the volumesnapshot case as I believe volumesnapshots won't go into bound state 
  until the pvc has proceeded to bound state.

**Related issues:**
https://github.com/backube/volsync/issues/246
https://github.com/backube/volsync/issues/48

### DIFF
--- a/controllers/mover/events.go
+++ b/controllers/mover/events.go
@@ -27,9 +27,6 @@ const (
 	// that a VolumeSnapshot object is not bound to a VolumeSnapshotContent
 	// object.
 	SnapshotBindTimeout = 30 * time.Second
-	// SnapshotReadyTimeout is the amount of time we should wait before warning
-	// that a VolumeSnapshot object is not ready.
-	SnapshotReadyTimeout = 120 * time.Second
 	// PVCBindTimeout is the time we should wait before warning that a PVC
 	// object is not bound to a PV.
 	PVCBindTimeout = 120 * time.Second
@@ -44,7 +41,6 @@ const (
 	EvRTransferFailed  = "TransferFailed" // Warning
 	EvRSnapCreated     = "VolumeSnapshotCreated"
 	EvRSnapNotBound    = "VolumeSnapshotNotBound" // Warning
-	EvRSnapNotReady    = "VolumeSnapshotNotReady" // Warning
 	EvRPVCCreated      = "PersistentVolumeClaimCreated"
 	EvRPVCNotBound     = "PersistentVolumeClaimNotBound" // Warning
 	EvRSvcAddress      = "ServiceAddressAssigned"
--- a/controllers/mover/events.go
+++ b/controllers/mover/events.go
@@ -27,6 +27,9 @@ const (
 	// that a VolumeSnapshot object is not bound to a VolumeSnapshotContent
 	// object.
 	SnapshotBindTimeout = 30 * time.Second
+	// SnapshotReadyTimeout is the amount of time we should wait before warning
+	// that a VolumeSnapshot object is not ready.
+	SnapshotReadyTimeout = 120 * time.Second
 	// PVCBindTimeout is the time we should wait before warning that a PVC
 	// object is not bound to a PV.
 	PVCBindTimeout = 120 * time.Second
@@ -41,6 +44,7 @@ const (
 	EvRTransferFailed  = "TransferFailed" // Warning
 	EvRSnapCreated     = "VolumeSnapshotCreated"
 	EvRSnapNotBound    = "VolumeSnapshotNotBound" // Warning
+	EvRSnapNotReady    = "VolumeSnapshotNotReady" // Warning
 	EvRPVCCreated      = "PersistentVolumeClaimCreated"
 	EvRPVCNotBound     = "PersistentVolumeClaimNotBound" // Warning
 	EvRSvcAddress      = "ServiceAddressAssigned"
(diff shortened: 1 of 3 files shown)


tesshuflower:
> > That is, bound but no restoreSize set yet
> > and then later get one with restore size set - so in this sort of situation we'll always be falling back to the PVC capacity.
> > We could potentially try to wait for the snapshot readyToUse: true before trying to create a PVC?
> 
> Turns out that restoreSize is optional. I found it documented here: https://kubernetes-csi.github.io/docs/api/volume-snapshot.html#snapshot.storage.k8s.io/v1.VolumeSnapshotStatus
> 
> I think waiting for `readyToUse` will probably fix our race. The caveat is that it can take a long time for a snap to become ready, particularly for cloud vendors. Explicitly waiting for it will just cause us to pause at a different part of the sequence.

Hey that doc is great, thx for that - Do you think it's safe to wait on `readyToUse`?  It seems that field is also optional

JohnStrunk:
I'm not aware of any drivers that don't report `readyToUse`, but optional does mean optional. I'm concerned about the test for nil though. How do we figure out "will never be reported, so ignore it" vs "haven't reported yet"?
[This comment in the code](https://github.com/kubernetes-csi/external-snapshotter/blob/de06dbc559f9ba4258768a0268be84518e0e2ea5/client/apis/volumesnapshot/v1/types.go#L121-L131) about eventual consistency worries me.

I wonder if we're going to have to rely on PVC resize or something... During reconcile figure out the size based on currently available info, and adjust the PVC request upward if necessary.

tesshuflower:
> I'm not aware of any drivers that don't report `readyToUse`, but optional does mean optional. I'm concerned about the test for nil though. How do we figure out "will never be reported, so ignore it" vs "haven't reported yet"? [This comment in the code](https://github.com/kubernetes-csi/external-snapshotter/blob/de06dbc559f9ba4258768a0268be84518e0e2ea5/client/apis/volumesnapshot/v1/types.go#L121-L131) about eventual consistency worries me.
> 
> I wonder if we're going to have to rely on PVC resize or something... During reconcile figure out the size based on currently available info, and adjust the PVC request upward if necessary.

I guess I was hoping that we could keep the bound check there, and only after it's bound, check for the readyToUse - this assumes that if `BoundVolumeSnapshotContentName` is set, then `readyToUse` is initialized and we can rely on the value (i.e. if the driver will set readyToUse, they'll initialize with true/false and then update to true at some point).  Not sure if this is a good assumption.

JohnStrunk:
> Edit: I've made a change here to attempt to do this: https://github.com/backube/volsync/commit/5559407405d11a58190e511678a408e579c4ff20. @JohnStrunk what do you think about doing it this way?

Seems like the best we can do.
As for the event in that commit... I don't think we want to publish an Event for not ready. The cloud providers can take a really long time to become ready in the normal case. I added the original event based on "bound" since that should be sufficient indication that the driver supports the operation and hasn't rejected it.

tesshuflower:
> > Edit: I've made a change here to attempt to do this: [5559407](https://github.com/backube/volsync/commit/5559407405d11a58190e511678a408e579c4ff20). @JohnStrunk what do you think about doing it this way?
> 
> Seems like the best we can do. As for the event in that commit... I don't think we want to publish an Event for not ready. The cloud providers can take a really long time to become ready in the normal case. I added the original event based on "bound" since that should be sufficient indication that the driver supports the operation and hasn't rejected it.

Makes sense, I'll back out the event.

If the snapshot can take a long time to become ready, we may start to see this sort of event too?
https://github.com/backube/volsync/pull/251/commits/5559407405d11a58190e511678a408e579c4ff20

I don't think the pvc will move into bound phase until the snapshot is ready

JohnStrunk:
Good point-- we may, indeed. Though w/ your new ready check, we should avoid it since we won't proceed from snapshot to PVC until it's ready.

tesshuflower:
I've rebased to pickup the latest updates in main

/cc @JohnStrunk  I think this is ready for review now

JohnStrunk (approved):

PR #242, opened 2022-05-31
Upgrade operator-sdk to 1.21.0

JohnStrunk:
**Describe what this PR does**
- Upgrades operator-sdk to 1.21.0

**Is there anything that requires special attention?**

**Related issues:**

PR #255, opened 2022-06-06
Fix Makefile prereqs

JohnStrunk:
When I upgraded the Makefile in #242, I didn't get the target dependencies correct. The result is that the build process attempts to download the helper tools every time they may be needed, even if they are already present in the `bin` directory.

**Related issues:**
Fix for issue introduced in #242


JohnStrunk:
/retest
Strange failure.

Issue #258, opened 2022-06-15
CSV needs to be updated for 0.4.1 release in the release-0.4 branch


tesshuflower:
To correspond with the ACM 2.5.1 zStream release, we'll need to version our CSV that ACM picks up.

Atm I don't think we have fixes destined for release-0.4.1 but we do need to version the CSV so that ACM 2.5.1 can publish a new version - this is because images will be re-spun to pickup fixes in the base ubi8 image (and possibly golang as well).

After this step is done, the ACM midstream will need to be modified to start picking up from the release-0.4 branch again instead of the hardcoded commit that corresponds with VolSync 0.4.0

tesshuflower:
Maybe this should have been labelled as a bug

tesshuflower:
Fixed by: https://github.com/backube/volsync/pull/260

PR #260, opened 2022-06-15
Version operator to 0.4.1

tesshuflower:

Updates the operator CSV metadata to 0.4.1

- Did not add an olm.SkipRange as we'll be publishing to the same acm-2.5 channel, so a .z update should automatically be applied within the channel.
- In version.mk I left the CHANNEL list to include `stable`.  We may need to make sure we remove stable once we publish `0.5.0`.  I.e. if we ever publish a `0.4.2` after `0.5.0` do we need to be careful not to publish `0.4.2` to the stable channel?
- Do any of these files need to be updated with the `0.4.1` version?
  - CHANGELOG.md
  - kubectl-volsync/volsync.yaml
  - helm/volsync/Chart.yaml

**Related issues:**
Fixes https://github.com/backube/volsync/issues/258

### DIFF
--- a/bundle/manifests/volsync.clusterserviceversion.yaml
+++ b/bundle/manifests/volsync.clusterserviceversion.yaml
@@ -43,7 +43,7 @@ metadata:
     capabilities: Basic Install
     operators.operatorframework.io/builder: operator-sdk-v1.14.0+git
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
-  name: volsync.v0.4.0
+  name: volsync.v0.4.1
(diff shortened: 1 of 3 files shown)


JohnStrunk:
> Do any of these files need to be updated with the 0.4.1 version?

- The Helm chart version should get updated (we should release a new version for 0.4.1)
- The CLI version should be updated automatically
- The CHANGELOG should get a section for 0.4.1, but that probably belongs in main instead of release-0.4

tesshuflower:
Think this is ready for review now

/cc @JohnStrunk

PR #269, opened 2022-06-17
Another try at fixing the `make` dependency chain

JohnStrunk:
`make` was still trying to repeatedly download tools during the build process. This PR removes the re-direction through all the phony targets, using the actual tool executables as dependencies.

**Related issues:**
Follow-on to #255


JohnStrunk:
/close
I suspect this is a losing proposition.

PR #272, opened 2022-06-17
Update nextSyncTime every time we reconcile while in "cleanup" 

JohnStrunk:
We were only updating `nextSyncTime` when we transitioned Syncing => Cleanup. This caused us to not update the NST if someone changed the schedule.
This fix re-calculates NST based on the last completion time every time we reconcile while in cleanup. Since LST is only updated when sync completes, this results in no change to NST in the normal case. However, if the schedule is changed, it will cause a reconcile, and that new cronspec will be used to calculate NST.

**Related issues:**
Fixes #217


tesshuflower (approved):
/lgtm

PR #274, opened 2022-06-17
Upgrade operator-sdk to v1.22.0

JohnStrunk:
Upgrades scaffolding to sdk 1.22

This looks like it's starting to scaffold some of the new security stuff that we're going to need to address soon.

**Related issues:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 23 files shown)


tesshuflower (commented):
I think we may need to make sure the crds in bundle/manifests are also updated to be in-sync with the ones in `crd/bases`.  Running `make bundle` will probably take care of this.

JohnStrunk:
Good catch

JohnStrunk:
/retest
That's a strange unit test failure (etcd timeout)

PR #284, opened 2022-06-22
Remove requirement for CAP_AUDIT_WRITE in rsync

### LINKED ISSUE #257: Investigate whether AUDIT_WRITE can be removed from rsync mover pod
Permissions will start to be an issue with Pod Security Admission updates coming in Kubernetes (which for example will be on in OCP 4.11, but perhaps only enforced in OCP 4.12?).

`sshd` in the rsync mover pod requires `AUDIT_WRITE` capability - we think this is mostly to be able to write audit logs (rsync doesn't seem to have any option to disable writing these).

Can we remove the AUDIT_WRITE capability from the rsync mover pod by mounting an `emptyDir` volume and mounting it to where rsync writes its audit logs?

- AUDIT_WRITE capability no longer required for rsync mover pod

JohnStrunk:
The AUDIT_WRITE capability was required due to pam_loginuid.so. This patch marks that pam module as optional so that failing it doesn't prevent sshd from working.

**Related issues:**
Fixes #257

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - "Reconciled" condition removed from ReplicationSource and
   ReplicationDestination `.status.conditions[]` in favor of returning errors via
   the "Synchronizing" Condition.
+- Removed the CAP_AUDIT_WRITE permission from the rsync mover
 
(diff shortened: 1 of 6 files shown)


JohnStrunk:
Fixes #257

JohnStrunk:
I'm at the IWFM stage of investigation. :laughing: 
Also looking into the CAP_CHROOT.

PR #297, opened 2022-06-30
Clear error state of Synchronizing condition after successful reconcile pass

### LINKED ISSUE #290: Returning a nil error doesn't clear past errors
**Describe the bug**

When a mover returns an error, VolSync sets the Synchronizing condition to false, e.g.:
```json
  {
    "lastTransitionTime": "2022-06-28T17:23:55Z",
    "message": "the peer list contains the node itself",
    "reason": "Error",
    "status": "False",
    "type": "Synchronizing"
  }
```
After subsequent successful syncs where `nil` is being returned in the error field, the Synchronizing condition remains stuck at `False` with the last non-nil error set as the message & transition time.

This happens when `Result.RetryAfter` is set, I'm not sure if setting `Result.Complete: true` clears it or not. 
 
**Steps to reproduce**

Check out PR #134 and run the operator with a Syncthing ReplicationSource, the reconcile condition will always be set to whatever the last error was.

**Expected behavior**

The synchronization status should be set to `InProgress` since the Syncthing synchronization is always happening.

**Actual results**

The synchronization status is set to `False` due to an error, with whatever the last error that occurred was.

**Additional context**

I've only observed this issue while developing the Syncthing ReplicationSource, but have not tried replicating this with other movers.

JohnStrunk:
The Synchronizing condition was only being updated when the state changed. This caused a problem if a reconcile pass returned a transient error. The error condition state would remain until the next Synchronizing/Cleanup state change instead of returning to the proper true/false condition state w/ an appropriate reason code.

This change ensures that the condition state gets updated w/ each reconcile pass

**Related issues:**
Fixes #290

### DIFF
--- a/controllers/statemachine/machine_test.go
+++ b/controllers/statemachine/machine_test.go
@@ -19,6 +19,7 @@ package statemachine
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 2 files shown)


controllers/statemachine/machine_test.go
```diff
@@ -280,3 +281,36 @@ var _ = Context("Issue 271: nextSyncTime not being updated", func() {
 		Expect(m.NextSyncTime().IsZero()).To(BeFalse())
 	})
 })
+
+var _ = Context("Issue 290: Synchronizing condition error doesn't clear", func() {
```
RobotSail, line 285: The practice of placing an issue# within testing is definitely something that I will have to adopt

RobotSail (approved):
Looks good to me. So the problem was just that the state wasn't being set when the result wasn't complete?

JohnStrunk:
> Looks good to me. So the problem was just that the state wasn't being set when the result wasn't complete?

Yeah. The status was being set when the state changed... because... that's when it needs to change. :smile:
But since the errors override the current status, that was a flawed assumption.

PR #298, opened 2022-07-01
Update bundle files from syncthing merge

JohnStrunk:
The bundle files didn't get updated with the Syncthing mover PR. This includes the updated files in the bundle.

**Related issues:**
Follow-on to #134

### DIFF
--- a/bundle/manifests/volsync.backube_replicationsources.yaml
+++ b/bundle/manifests/volsync.backube_replicationsources.yaml
@@ -130,8 +130,8 @@ spec:
                     minItems: 1
                     type: array
                   cacheAccessModes:
-                    description: accessModes can be used to set the accessModes of
-                      restic metadata cache volume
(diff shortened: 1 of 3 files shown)


tesshuflower (commented):
@JohnStrunk I think we'll need to inject the syncthing env var into the deployment yaml for the operator as well (and then `make bundle` should pull it all into the csv).  I forgot this is how the operator sets the images to be used for the movers, using env vars rather than cmd line flags.

in config/manager/kustomization.yaml we'll need to add something like this:

```
- patch: |-
    - op: add
      path: /spec/template/spec/containers/0/env/-
      value:
        name: RELATED_IMAGE_SYNCTHING_CONTAINER
        value: quay.io/backube/volsync-mover-syncthing:latest
  target:
    kind: Deployment
```

tesshuflower (approved):
/lgtm

Issue #291, opened 2022-06-29
Upgrading VolSync when a job is running - job reconciles can fail repeatedly


tesshuflower:
**Describe the bug**
When VolSync is updated (seems to happen from 0.4.x to the newer 0.5.0/main builds), if an existing job is running, VolSync will attempt to reconcile and get an error:

```
[...]

PR #302, opened 2022-07-04
Delete/recreate jobs if immutable field needs update

tesshuflower:
For issue https://github.com/backube/volsync/issues/291

- Adds a util to do a CreateOrUpdate on a resource but also catch update errors about immutable fields and delete the resource (so it can be recreated on next reconcile).

- Uses the new util for mover jobs.  This way if immutable fields (one common example may be the mover job image in the job pod) need update, the job can be deleted and recreated.

**Related issues:**
https://github.com/backube/volsync/issues/291

### DIFF
--- a/controllers/mover/rclone/mover.go
+++ b/controllers/mover/rclone/mover.go
@@ -29,7 +29,6 @@ import (
 	"k8s.io/client-go/tools/events"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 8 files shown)


tesshuflower:
/cc @JohnStrunk 

Not sure what to say about the sonarcloud code analysis - it's complaining about 2 lines of duplicated code I think?  one is returning err, nil and the other is calling a function.

JohnStrunk:
/lgtm
In general, I have been unimpressed by SonarCloud.

PR #307, opened 2022-07-07
Enable periodic builds of release branches

### LINKED ISSUE #266: CI: Scheduled runs for release branches
We have a weekly scheduled job that runs against `main`, but we don't have the equivalent for the release branches... We should.

By only running CI when something gets backported, we end up w/ bit rot on the release branches. This slows our ability to publish fixes to z-streams.

- "Active" release branches have scheduled CI runs
- Release branches can be easily moved between active and inactive
- The set of tests and environment can be different between branches

**Additional context**
Example:
- Yesterday, #260 was a simple PR to bump the version 0.4.0 -> 0.4.1 in `release-0.4` branch, but because we hadn't backported the CI fix from #249, CI on the release branch was broken and we didn't find out about it until we tried to make a change.

I think this will require some GH action magic because scheduled runs are only supported on the default branch. We can hopefully get around it by [having one workflow trigger another](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch).

JohnStrunk:
- Enable triggering docs and operator flows manually. This is needed for upcoming scheduled builds
- Consolidate the container push jobs and turn them into a re-usable action to cut down on code duplication

**Related issues:**
Fixes #266


JohnStrunk:
/cc @tesshuflower 
There's a lot of change to the workflows here. I did my best to test things in a separate repo, but it's still :crossed_fingers:

JohnStrunk:
/retest
Due to #309

tesshuflower (approved):
/lgtm

PR #315, opened 2022-07-11
CI: Fix env substitution when calling push workflow

JohnStrunk:
#307 changed the workflow for pushing container images, and broke it.
The container image names are stored as env variables, but the normal `${FOO}` syntax doesn't work for setting inputs to other workflows. It turns out that the `env` context also isn't available at that point in the workflow file, so I have resorted to just putting the value there.

From the action log, we see that the variable reference didn't get substituted. :disappointed: ... well, it did get substituted, it just needed to be done before the workflow call, not after.
```
Run docker load -i /tmp/image.tar
  docker load -i /tmp/image.tar
  docker inspect ${OPERATOR_IMAGE}
  shell: /usr/bin/bash -e {0}
Loaded image: quay.io/backube/volsync:latest
"docker inspect" requires at least 1 argument.
See 'docker inspect --help'.
Usage:  docker inspect [OPTIONS] NAME|ID [NAME|ID...]
Return low-level information on Docker objects
Error: Process completed with exit code 1.
```

**Related issues:**

### DIFF
--- a/.github/workflows/operator.yml
+++ b/.github/workflows/operator.yml
@@ -400,13 +400,16 @@ jobs:
       - name: No-op
         run: /bin/true
 
+  # The operator images are specified directly because the env context isn't
+  # available in the job.with.
+  # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idwithinput_id
   push-operator:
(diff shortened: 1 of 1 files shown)


tesshuflower (approved):
/lgtm

Issue #305, opened 2022-07-06
CI flake: Should remove old snapshot  test in rclone_test fails intermittently


tesshuflower:
**Describe the bug**

```
Failure [30.170 seconds]
Rclone as a destination when used as destination Cleanup is handled properly [It] Should remove old snapshot 
/home/runner/work/volsync/volsync/controllers/mover/rclone/rclone_test.go:1009
Timed out after 30.001s.
  Expected
      <int>: 2
  to equal
      <int>: 1
  /home/runner/work/volsync/volsync/controllers/mover/rclone/rclone_test.go:1018
------------------------------
[BeforeEach] Rclone as a destination
  /home/runner/work/volsync/volsync/controllers/mover/rclone/rclone_test.go:721
[BeforeEach] Cleanup is handled properly
  /home/runner/work/volsync/volsync/controllers/mover/rclone/rclone_test.go:906
[JustBeforeEach] Rclone as a destination
  /home/runner/work/volsync/volsync/controllers/mover/rclone/rclone_test.go:743
[JustBeforeEach] when used as destination
  /home/runner/work/volsync/volsync/controllers/mover/rclone/rclone_test.go:751
[JustBeforeEach] Cleanup is handled properly
  /home/runner/work/volsync/volsync/controllers/mover/rclone/rclone_test.go:940
[It] Should remove old snapshot
  /home/runner/work/volsync/volsync/controllers/mover/rclone/rclone_test.go:1009
1.657[122](https://github.com/backube/volsync/runs/7218145427?check_suite_focus=true#step:5:123)4033852162e+09	DEBUG	Starting cleanup	{"method": "Rclone", "m.mainPVCName": "dest", "m.isSource": false}
1.6571224033852265e+09	DEBUG	removing snapshot annotations from pvc	{"method": "Rclone"}
1.6571224033878293e+09	INFO	deleting temporary objects	{"method": "Rclone", "owned-by": "7c8f8994-2ea1-4c41-9ad9-5dd14f762a9f"}
1.6571224033908823e+09	DEBUG	Cleanup complete	{"method": "Rclone"}
[AfterEach] Rclone as a destination
  /home/runner/work/volsync/volsync/controllers/mover/rclone/rclone_test.go:746
Summarizing 1 Failure:
[Fail] Rclone as a destination when used as destination Cleanup is handled properly [It] Should remove old snapshot 
/home/runner/work/volsync/volsync/controllers/mover/rclone/rclone_test.go:1018
```

[...]

Issue #267, opened 2022-06-16
CI Flake: Wrap `client.Create()` calls


JohnStrunk:
The below needs to be wrapped to ensure the object is visible:
https://github.com/backube/volsync/blob/1d00a461beb074263a394a1bb89d25096cd8dc10/controllers/mover/rsync/rsync_test.go#L1143

```
Rsync as a destination when used as destination SSH keys are handled properly When ssh keys are provided When provided secret exists but missing fields [It] Mover should fail to ensureSecrets 
/home/runner/work/volsync/volsync/controllers/mover/rsync/rsync_test.go:1148
Expected
      <string>: Secret "test-dest-keys-bad" not found
  to contain substring
      <string>: fields
  /home/runner/work/volsync/volsync/controllers/mover/rsync/rsync_test.go:1152
```

We should probably create a helper method to replace the `client.Create()` call to handle these cases, and do a search/replace.

tesshuflower:
/assign @tesshuflower 

I can take a look at this - I have another similar issue: https://github.com/backube/volsync/issues/291. Where a status().Update() is also not getting picked up by the cache sometimes - maybe this can be wrapped as well.

tesshuflower:
/close

closed by https://github.com/backube/volsync/pull/316

PR #316, opened 2022-07-11
Unit test flakes client cache

tesshuflower:
- adds a helper function for unit tests to create an object and also wait for it to be loaded into the k8s client cache.  This is for unit tests that are using a reader cache and have objects that need to exist prior to the rest of the test running (creating a namespace for example)
- Moves some unit tests that aren't relying on controllers to use a direct k8s client  - the client that reads directly from the cache - to avoid these cache issues altogether

**Related issues:**
https://github.com/backube/volsync/issues/267
https://github.com/backube/volsync/issues/305

### DIFF
--- a/controllers/mover/rclone/rclone_test.go
+++ b/controllers/mover/rclone/rclone_test.go
@@ -230,6 +230,7 @@ var _ = Describe("Rclone as a source", func() {
 			},
 		}
 		Expect(k8sClient.Create(ctx, ns)).To(Succeed())
+
 		Expect(ns.Name).NotTo(BeEmpty())
 
 		sc := "spvcsc"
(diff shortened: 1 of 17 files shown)


tesshuflower:
@JohnStrunk Would you mind taking a look and seeing if you'd be ok with this approach?

Essentially what I've done is to create a `createWithCacheReload()` function for tests using a caching k8s client that will create the object and wait for it to appear with the caching client's GET().

However while looking at this I also realized that a lot of our tests are using the caching client when maybe we don't need to?

For the moment I've updated the controllers/mover/rclone tests to use a direct client - what do you think of this approach?  I think really the place where we benefit from the caching client are tests that start controllers as this probably simulates a "real" environment more closely.  

I think we could probably leave the controllers tests using the caching client but move most of the other tests to a direct client that reads/writes directly to the API server.  This way we'd remove a lot of timing issues we see in our tests and remove the need for a lot of Eventually() functions.

JohnStrunk:
I think it's important to make sure the controllers themselves keep using a cache, but for direct manipulation in the test code, it should be fine to use a direct client. We don't care about the load on the test api server.

Issue #319, opened 2022-07-15
Support backing up to NFS/existing PVC using restic


onedr0p:
**Describe the feature you'd like to have.**

Support backing up to an PVC using restic

**What is the value to the end user? (why is it a priority?)**

Instead of object storage Restic could back up to an existing PVC (like one that is a NFS mount)

```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  name: volsync-nfs
spec:
  storageClassName: volsync-nfs
  capacity:
    storage: 1Mi
  accessModes:
    - ReadWriteMany
  persistentVolumeReclaimPolicy: Retain
  nfs:
    server: "192.168.1.1"
    path: /tank/backups
  mountOptions:
    - nfsvers=4.2
    - nconnect=8
    - hard
    - noatime
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: volsync-nfs
  namespace: home
spec:
  accessModes:
    - ReadWriteMany
  storageClassName: volsync-nfs
  resources:
    requests:
      storage: 1Mi
```

Maybe you could add a option in the restic config, e.g.

```yaml
restic:
  # If `RESTIC_REPOSITORY` is a local path, backup to an existing PVC
  existingPVC: volsync-nfs
```

But maybe there's a better way to do this, like embedding the nfs options into the config?

```yaml
restic:
  # Back up to an NFS server
  nfs:
    server: 192.168.1.1
    path: /tank/backup/app
```

**How will we know we have a good solution? (acceptance criteria)**

- Back up to NFS Server and/or `existingClaim`

**Additional context**

onedr0p:
While I can see `moverVolumes` was added it only accepts an existing claim or a secret. There is no option to use an inline nfs option.

Ideally it would be nice if this worked or something like it.

```yaml
    moverVolumes:
      - mountPath: /repo
        volumeSource:
          nfs:
            path: /mnt/eros/Media
            server: expanse.internal
```

The reason for this is that PVC is namespace scoped, and each namespace would need an exiting claim pointing to a NFS PV (see issue description)

tesshuflower:
Makes sense - I should be able to add support for nfs volumeSources pretty easily - we initially only implemented secret and PVC type as you've seen.

onedr0p:
This was completed

Issue #320, opened 2022-07-15
Support a kopia-based backup


onedr0p:
**Describe the feature you'd like to have.**

Hi again 👋🏼 

Would it be possible to support Kopia? I know you already have support for restic, rclone and rsync, but I figured Kopia would fit in nicely as it's becoming a very popular tool for incremental data backup. In my tests, it is also very fast.

https://github.com/kopia/kopia

It is written in Go and has decent APIs

https://pkg.go.dev/github.com/kopia/kopia/repo?utm_source=godoc

**What is the value to the end user? (why is it a priority?)**

Supporting another way of backing and restoring data.

**How will we know we have a good solution? (acceptance criteria)**

**Additional context**

onedr0p:
Yeah those benchmarks are pretty clear that Kopia is a serious contender to restic :)

trieded:
hi any plans for this?

onedr0p:
There's needs to be some considerations made on how to implement this since kopia works a little different than existing movers. For example kopia can support a single "kopia repository" for multiple apps and "kopia snapshots" are the way you backup or restore data.

onedr0p:
#  Local Filesystem Commands

This method is for using Kopia on your filesystem, not backing up to s3. Kopia does support s3 but I need to look into how that works. It's probably quite similar.

- `/snapshots` is the location of the Kopia Repository
- `/data` is the location of the PVC root filesystem

Note that in the examples below `/snapshots` is a single Kopia Repository with multiple snapshots across all PVCs that are backed up. This can help with deduplication if all the PVC volumes are in the same Kopia Repository. It could also be architected to support a 1:1 snapshot to Kopia Repository like the restic implementation is currently doing.

At a glance in the Kopia web UI, a single Kopia Repository with multiple snapshots (PVCs) looks like this:

<img width="1289" alt="image" src="https://user-images.githubusercontent.com/213795/204154483-cf38f274-e343-4770-b942-61130e864466.png">

## Create backup

```bash
[[ ! -f /snapshots/kopia.repository.f ]] && kopia repository create filesystem --path=/snapshots
kopia repo connect filesystem --path=/snapshots --override-hostname=cluster --override-username=root
kopia policy set /data/<namespace/<app>/<claim> --compression=zstd --keep-latest 14 --keep-hourly 0 --keep-daily 7 --keep-weekly 2 --keep-monthly 0 --keep-annual 0
kopia snap create /data/<namespace/<app>/<claim>
kopia repo disconnect
[...]

PR #321, opened 2022-07-15
feat: add source and destination support for nfs volumes

onedr0p:
Signed-off-by: Devin Buhl <devin@buhl.casa>

Adds support to mount a NFS volume as the data volume in the data mover pods.

I have implemented this in all of the movers, let me know what you all think and I can update the docs on the rest of the movers.

I have not fully tested this other than it building and running. I am interested in feedback. One thing to observe is that the `/data` mount point in the data mover pods is a "local" volume, so should we be overriding the `RESTIC_REPOSITORY` to `/data` when the nfs options are set in the config?

**Related issues:**

https://github.com/backube/volsync/issues/319


JohnStrunk:
@onedr0p Thanks for filing the issue and posting a PR.
Some thoughts:
- Since this is only applicable to the restic mover, I'd like to limit the changes to that portion of the code.
- I'd also like to have this be more generic by just referencing a PVC name instead of NFS directly.

Based on the above, you could add the following to `ReplicationDestinationResticSpec`
```go
	// repositoryPVC is the name of an existing PVC containing the backup repository
	//+optional
	RepositoryPVC string `json:"repositoryPVC,omitempty"`
```

Then in the restic mover controller, if RepostoryPVC is filled in, it can be added as a volume mount on `/repository`, allowing the restic Secret to refer to that "local" path.

onedr0p:
Hi @JohnStrunk thanks for the feedback, I agree with you on all points.

My only thought would be could it be possible to use the same NFS PVC for all applications? For example say I have 3 applications I want backed up using restic, could I reuse the same PVC and just point each to `/repository/<app-name>` in the CR?

Small background, I was able to use [Kyverno and have it create cronjobs for backing up PVCs with a certain label to a Kopia repository](https://github.com/onedr0p/home-ops/blob/main/cluster/apps/kyverno/policies/snapshot-cronjob-controller.yaml). After doing this, I discovered volsync and really loved the simplicity of this project and would love to integrate Kopia #320 as a data mover here as well. I was thinking too fork volsync and have it match my usecase but I'm pretty happy with your progress and diligence so far.

JohnStrunk:
Since the (NFS) PVC would be RWX, use by multiple (mover) pods simultaneously would be fine. You'd just need to have a unique restic Secret per ReplicationSource that holds the sub-path (`/repository/subpath`) for each source that you want to back up.

Glad you find the project useful. :heart:

onedr0p:
Closing this PR as I haven't had time to get back to it. Maybe someday I can get back to hacking at it.

PR #331, opened 2022-07-27
Update VolSync description in csv

tesshuflower:

Updates the volsync description in the CSV.

- Updated the csv in config/manifest/bases and then ran `make bundle`.

- Will need to cherry-pick into release-0.5

### DIFF
--- a/config/manifests/bases/volsync.clusterserviceversion.yaml
+++ b/config/manifests/bases/volsync.clusterserviceversion.yaml
@@ -22,7 +22,32 @@ spec:
       kind: ReplicationSource
       name: replicationsources.volsync.backube
       version: v1alpha1
-  description: Asynchronous volume replication for Kubernetes CSI storage
+  description: "Asynchronous volume replication for Kubernetes CSI storage\n\n VolSync
+    is a Kubernetes operator that performs asynchronous replication of\npersistent
+    volumes within, or across, clusters. VolSync supports replication in a\nstorage
+    system independent manner. This means replication can be used with\nstorage systems
+    that do not support replication natively. Data can also be\nreplicated across
+    different types (and vendors) of storage.\n\nVolSync supports both 1:1 replication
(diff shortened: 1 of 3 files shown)


config/manifests/bases/volsync.clusterserviceversion.yaml
```diff
@@ -22,7 +22,32 @@ spec:
       kind: ReplicationSource
       name: replicationsources.volsync.backube
       version: v1alpha1
-  description: Asynchronous volume replication for Kubernetes CSI storage
+  description: "Asynchronous volume replication for Kubernetes CSI storage\n\n VolSync
```
JohnStrunk, line 25: Can we use the ": |" yaml functionality to make this a bit easier to read?
[SnapScheduler example](https://github.com/backube/snapscheduler/blob/a01f63d35ef1a9ee6362706f905d9c235d1b3feb/config/manifests/bases/snapscheduler.clusterserviceversion.yaml#L101-L172)

tesshuflower: I actually originally inserted it in that format you suggested, but as soon as you run `make bundle`, it re-formats it.  Not sure why it doesn't only format the one in bundle/manifests.

I just tried a test with a simpler description and it does preserve the ": |" format.  There must be something in there (I'm guessing length?) that causes it to decide to re-format. I think it's the `operator-sdk generate bundle` step that's doing it.

JohnStrunk: oh, that's really strange. I'm pretty sure the flow is the same w/ SnapScheduler and it doesn't cause a problem.

tesshuflower: Update: it turned out to be trailing spaces in the bullet points - meant to indicate a line break.  During the yaml conversion this caused it to get reformatted - I've tried again but had to add newlines in-between the bullet points subject and text.

PR #335, opened 2022-07-28
tests: Switch tests from kuttl to Ansible

JohnStrunk:
Since we've been bumping against the limits of kuttl for some time, this is an exploration of moving the e2e tests into Ansible instead.

- This is an experiment... I'd like feedback from others.
  - Does this look awful? Does it look impossible to write?
  - Does it look any better than kuttl?
  - *I happen to think this is really promising... the tests are largely straight-line, and roles give us the ability to re-use code.*

About the structure:
My intent is that each test would have a top-level `test_foo.xml` that is a playbook. That playbook would have the test, plus callouts into individual roles where there is re-usable code. A good example is the role "create_namespace". It creates a randomly named Namespace to be used for a test, and sets up a handler to delete the NS at the test's conclusion.

One thing that kuttl did for us in the past is to run tests in parallel. This isn't possible w/ Ansible... it runs a single playbook on multiple hosts in parallel, not multiple playbooks on a single host (in parallel). there is, however `ansible-parallel` that does. This will get us kuttl's parallelism.

[Quickstart in the README.md](https://github.com/JohnStrunk/volsync/tree/ansible/test-e2e/README.md)

**Related issues:**

### DIFF
--- a/.ci-scripts/pre-commit.sh
+++ b/.ci-scripts/pre-commit.sh
@@ -22,7 +22,8 @@ function run_check() {
         find . \
             -path ./vendor -prune -o \
             -path ./.venv -prune -o \
-            -path ./testbin -prune -o \
+            -path ./test-e2e/.collections -prune -o \
+            -path ./test-e2e/.venv -prune -o \
             -regextype egrep -iregex "$regex" -print0 | \
             xargs -0rt "$exe" "$@"
         echo
--- a/.ci-scripts/yamlconfig.yaml
+++ b/.ci-scripts/yamlconfig.yaml
@@ -15,3 +15,4 @@ rules:
     ignore: |
       kubectl-volsync/volsync.yaml
       test-kuttl/**
+      test-e2e/**
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,8 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+
+  - package-ecosystem: "pip"
+    directory: "/test-e2e"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/operator.yml
+++ b/.github/workflows/operator.yml
@@ -372,6 +372,24 @@ jobs:
       - name: Run e2e tests
         run: make test-e2e
 
+      - name: Setup Python
+        uses: actions/setup-python@c4e89fac7e8767b327bbad6cb4d859eda999cf08
+        with:
+          python-version: '3.10'
+          cache: 'pipenv'
+
+      - name: Install pipenv
+        run: python -m pip install pipenv
+
+      - name: Initialize virtual env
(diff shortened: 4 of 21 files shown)


JohnStrunk:
@tesshuflower @RobotSail @pranavgaikwad @ShyamsundarR
Interested in your thoughts. All the better if you've never written Ansible before.

RobotSail:
At a surface level, this seems like a great way to simplify complex operations that will sometimes need to be broken out into bash code. I like that variables are explicitly defined across tests and can be included where needed, that is a big thing in the Syncthing tests especially because there were so many moving parts. 

Right now we're looking at writing end-to-end testing for the IPFS Cluster Operator, which requires a similar level of configuration and interaction between running resources as Syncthing, so I'd love to see how those tests could be written with Ansible.

tesshuflower:
I haven't used ansible before, but this does look promising - I also like the idea of the templates, shared "functions" and ability to update parameters in the k8s objects.  I didn't find kuttl to be all that easy to use/maintain either, so this doesn't seem worse to me in any way (at least at first).

@JohnStrunk From your experience does ansible get hard to maintain? I had the feeling you were reluctant to use it at first.

I've been trying to ask around (and am still waiting on some answers) about what people are generally using in ACM for e2e testing - there's a lot of tests that are using `cypress` which is essentially a javascript test framework mostly for UI testing.  I have the feeling it's used by a lot of the test team as they're familiar with it and it can do UI (and then they also use it for k8s/api testing).  I don't really see where it helps with any of the setup type tasks either.  I have heard (but not found any yet) that some are also using ginkgo directly - again, I think the issue here is we need some automation around the test setup/configuration primarily correct?

I'd be up to help try to migrate some of the existing kuttl tests over btw - this would help me get a hands-on idea of what is involved.

JohnStrunk:
> From your experience does ansible get hard to maintain? I had the feeling you were reluctant to use it at first.

My previous experience was in using it for configuring servers.
- I found "data manipulation" to be difficult/non-intuitive. Had to dynamically select subsets of servers for certain commands, but not others, etc.
- Testing the playbooks was hard to do, and getting it wrong could lead to destruction of the production servers. We don't have that problem here.

The directory structure also has quite a bit of sprawl as well... Want to create a role? Even if it's just 2 tasks, gotta put it in the roles directory as a subdir with a tasks dir under that and a main.yml.

I've actually been pleasantly surprised by how easy it's been so far for writing the tests.

JohnStrunk:
/cc @tesshuflower 
After thinking about it over the weekend, I think this is worth moving forward.
Opening this up for review and merge to let others take a stab at writing tests.

One thing I think is still missing is better log output on test failure. I think we can do this by wrapping the tests in a `block` and using `recover` that triggers a role to dump logs from the appropriate containers, etc. I'll work on another PR that adds that.

tesshuflower (approved):
/lgtm

PR #347, opened 2022-08-03
tests: Templatize Pods so they run as non-root on kube and OCP

JohnStrunk:
This turns the Pod descriptions for "write data" and "compare PVCs" roles into templates that get expanded differently based on the version of kube/openshift they are invoked against. This permits us to run the tests as non-uid0.

**Related issues:**


JohnStrunk:
/hold
I haven't tested this yet

tesshuflower (approved):
/lgtm

PR #351, opened 2022-08-05
tests: Set ansible tempdir to /tmp

JohnStrunk:
Changes the ansible config to put temporary files in /tmp instead of under $HOME. This appears to be necessary for openshift CI as $HOME doesn't seem to be writable.

**Related issues:**


tesshuflower (approved):
/lgtm

PR #353, opened 2022-08-05
Port restic e2e tests to ansible

tesshuflower:

Ports over the following kuttl tests over to ansible e2e test:
  -  `restic-with-manual-trigger`
  -  `restic-with-previous`
  - `restic-with-restoreasof`
  - `restic-without-trigger`

- I modified the `write_to_pvc` role to optionally leave the pod behind to allow for the podAffinity parts to work

- I had to alter the behavior of `restic-without-trigger` slightly as I wasn't able to catch the condition `Synchronizing` going to "false" - previously we did this in 20-ensure-multiple-syncs:
  ```
  kubectl -n $NAMESPACE wait --for=condition=Synchronizing=true --timeout=5m ReplicationSource/source
  kubectl -n $NAMESPACE wait --for=condition=Synchronizing=false --timeout=5m ReplicationSource/source
  kubectl -n $NAMESPACE wait --for=condition=Synchronizing=true --timeout=5m ReplicationSource/source
  ```

  I think with no trigger the status is changing so fast I wasn't able to catch it with the ansible k8s calls.  Instead I wait until
  another sync completes and make sure the lastSyncTime doesn't match the previous.  Might make the test take a bit longer .

- Additionally moves some of the pvc writer roles to use jobs instead of creating pods directly - this is to avoid issues when running tests manually with users with different permissions (for example running the tests as an admin user means the pods get created with a different SCC which affects the pod options).  It will still make sense to move the other pod roles (reader ones) into jobs, but this can be done in a separate PR.

**Related issues:**


tesshuflower:
@JohnStrunk I can continue with porting over the other restic kuttl tests - probably the 1st one was the most painful? 😃 

- I think for other restic tests I'll need to do something to make sure they don't all write to the same MinIO location

- Parts of it were easy where I could use the template, others I found a bit frustrating (i.e. ansible/jinja2 variables and default values and things like that - however this is probably just the learning curve as a new ansible user).

JohnStrunk:
> * I think for other restic tests I'll need to do something to make sure they don't all write to the same MinIO location

Adding a variable to use as the sub-path in the restic secret [here](https://github.com/backube/volsync/pull/353/files#diff-b7284206563b327442f0ff360f448cb33dc505f7c568fb02e5477a1204dbe78fR27) will probably do it.

----

Regarding the affinity setting, I just had to do that w/ the CLI. Instead of actually setting the affinity, I [ran the compare role on the 2 empty PVCs](https://github.com/backube/volsync/pull/354/files#diff-6c160dfe697496d3a1b46d2b5229642218e21849cc9af584ec38f8dede6b2a42R42-R49).

tesshuflower:
> > * I think for other restic tests I'll need to do something to make sure they don't all write to the same MinIO location
> 
> Adding a variable to use as the sub-path in the restic secret [here](https://github.com/backube/volsync/pull/353/files#diff-b7284206563b327442f0ff360f448cb33dc505f7c568fb02e5477a1204dbe78fR27) will probably do it.
> 

👍 Thx - was considering actually using the namespace in the path there so each test will get its own unique path

JohnStrunk:
Looks like you got hit by #359. Rebasing should fix the e2e
```
test_replication_sync_direct.yml>	TASK [cli : Execute cli command] ***********************************************
test_replication_sync_direct.yml>	Monday 08 August 2022  21:26:36 +0000 (0:00:00.023)       0:00:57.898 ********* 
test_replication_sync_direct.yml>	fatal: [localhost]: FAILED! => {"changed": false, "msg": "The ansible.builtin.command action failed to execute in the expected time frame (60) and was terminated"}
test_replication_sync_direct.yml>	
test_replication_sync_direct.yml>	NO MORE HOSTS LEFT *************************************************************
test_replication_sync_direct.yml>	
test_replication_sync_direct.yml>	PLAY RECAP *********************************************************************
test_replication_sync_direct.yml>	localhost                  : ok=22   changed=12   unreachable=0    failed=1    skipped=12   rescued=0    ignored=0   
```

tesshuflower:
latest commit will properly test affinity in restic e2e - expect e2e in this PR to fail until https://github.com/backube/volsync/pull/362 is merged.

tesshuflower:
/cc @JohnStrunk This may finally be ready to review 

- Note the tests seem to have been flaky lately, I've had to re-run a bunch of times - latest failure was another kuttl one, and I also had an issue previously I believe with creating a pvc from snapshot - I think this may be related to the one you were going to tune the timeouts on?

test-e2e/roles/pvc_affinity_pod/templates/pod.yml.j2
```diff
@@ -13,11 +13,7 @@ spec:
       command: ["sh", "-c"]
       args:
         - |
-          set -x -e -o pipefail
-
-          mkdir -p `dirname /mnt/{{ path }}`
-          echo '{{ data }}' > '/mnt/{{ path }}'
-          sync
+          sleep 99999
```
JohnStrunk, line 16: This Pod should be able to immediately exit. Once the volume has been provisioned, it's zone is locked.

test-e2e/test_restic_with_restoreasof.yml
```diff
@@ -0,0 +1,348 @@
...
+
+    - name: Save name of first snapshot
+      ansible.builtin.set_fact:
+        current_timestamp: "{{ current_timestamp_shell.stdout }}"
+
+    - name: Run pvc affinity pod for dest PVC for current restoreasof
+      include_role:
+        name: pvc_affinity_pod
+      vars:
+        pvc_names:
```
JohnStrunk, line 182: With only 1 PVC, I don't think this is setting a specific zone affinity.

JohnStrunk (commented):
Since the affinity role is just starting a pod that mounts multiple PVCs, what do you think of just using the pvc compare role instead? Alternatively, we could keep it and just have it exit/cleanup immediately. I don't think we're getting anything from having it leave the pod around.

test-e2e/roles/pvc_affinity_pod/templates/pod.yml.j2
tesshuflower: I just may not be understanding this all correctly - but I thought the idea behind leaving the pod running was to force later replicationsource/destintations that use direct to pull the node from that running pod.  In this way we confirm that the nodeSelector is working on the subsequent job?

JohnStrunk: :man_facepalming: Nope, I missed the point. I thought this role was for PVC provisioning as opposed to testing VolSync's RWO affinity.

PR #358, opened 2022-08-08
CI: Remove socat test pod from csi-hostpath in kind

JohnStrunk:
This removes the socat container before deploying the CSI hostpath driver.

The CSI hostpath driver exposes the `csi.sock` via TCP using a socat container. This is unnecessary for our usage--- it's there for testing the CSI driver.
It periodically causes problems for local kind clusters since it's pulled from docker.io and sometimes triggers throttling due to the number of pull requests coming from a given IP.

**Related issues:**


JohnStrunk:
Fix for test failure is in #359

tesshuflower (approved):
/lgtm

PR #359, opened 2022-08-08
tests: Allow overriding the command timeout for cli commands

JohnStrunk:
When I wrote the CLI role, I didn't consider that the role timeout wouldn't propagate to the actual CLI-calling task. This permits sending a parameter to the cli role that will be used to override the default task timeout when invoking the CLI executable

**Related issues:**


tesshuflower (approved):
/lgtm

PR #360, opened 2022-08-10
tests: Increase timeout waiting for syncthing pods to be deleted

JohnStrunk:
We're seeing cases where the syncthing pods aren't being deleted within the timeframe allowed in the test. Manual testing indicates that the pods do eventually get deleted. This appears to be an issue of scheduling/deleting latency on the kube-side.
This increases the timeout during the test.

**Related issues:**


tesshuflower (approved):
/lgtm

Issue #361, opened 2022-08-10
Restic mover stuck in pending, can't attach PVC and PVC not bound


JohnStrunk:
**Describe the bug**
- The Restic mover is stuck in pending
- The cache PVC is not bound and waiting for 1st consumer
- The cache PVC can't be attached because it's not bound

**Steps to reproduce**
- Restic mover
  - CopyMethod: Direct
  - Source PVC is RWO
- An existing pod is using the source PVC

**Expected behavior**
The mover pod should start

**Actual results**

**Additional context**
https://issues.redhat.com/browse/OADP-647

This appears to be due to the affinity setter which directly sets the nodeName in the podspec. This seems to completely bypass the scheduler, and subsequently bypass the volume provisioning step, leaving us w/ an unbound PVC that can't be attached.
This will only happen on the 1st invocation of the mover if an existing pod is using the sourcePVC (triggering the affinity setter). On subsequent invocations, the PVC will already be bound, so there should be no scheduling problems.

JohnStrunk:
This will need a backport to release-0.5 branch

tesshuflower:
/close

Fix merged into main and also cherry-picked into `release-0.5`.

PR #362, opened 2022-08-10
Use nodeSelector rather than nodeName to not bypass scheduler

tesshuflower:

Stops setting NodeName in the mover job spec and instead uses NodeSelector.  It seems that specifying NodeName directly bypasses the scheduler which means that if we have a PVC (like the restic cache pvc) that is in pending (because the storageclass has volumeBindingMode: `WaitForFirstConsumer`) - then the PVC will be stuck waiting for first consumer, and at the same time the mover pod is stuck waiting for the cache PVC to be Bound.  It seems using NodeSelector does go through the scheduler and everything starts as it should.

We are using the common node label `kubernetes.io/hostname` for the NodeSelector.

**Related issues:**
https://github.com/backube/volsync/issues/361#issuecomment-1211065869

### DIFF
--- a/controllers/mover/rclone/mover.go
+++ b/controllers/mover/rclone/mover.go
@@ -273,7 +273,7 @@ func (m *Mover) ensureJob(ctx context.Context, dataPVC *corev1.PersistentVolumeC
 				logger.Error(err, "unable to determine proper affinity", "PVC", client.ObjectKeyFromObject(dataPVC))
 				return err
 			}
-			job.Spec.Template.Spec.NodeName = affinity.NodeName
+			job.Spec.Template.Spec.NodeSelector = affinity.NodeSelector
 			job.Spec.Template.Spec.Tolerations = affinity.Tolerations
 		}
 		logger.V(1).Info("Job has PVC", "PVC", dataPVC, "DS", dataPVC.Spec.DataSource)
--- a/controllers/mover/restic/mover.go
+++ b/controllers/mover/restic/mover.go
@@ -393,7 +393,7 @@ func (m *Mover) ensureJob(ctx context.Context, cachePVC *corev1.PersistentVolume
 				logger.Error(err, "unable to determine proper affinity", "PVC", client.ObjectKeyFromObject(dataPVC))
 				return err
 			}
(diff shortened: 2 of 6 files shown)


tesshuflower:
Note: will update e2e tests to cover this scenario in PR https://github.com/backube/volsync/pull/353 to keep it separate from this change, which will need to be cherry-picked into `release-0.5`.  The e2e tests have changed since 0.5 so cannot be backported easily.

shubham-pampattiwar:
Thank you @tesshuflower for the fix !

tesshuflower:
Removed WIP - finished testing with affinity on/off, so I think this one is ok to merge now.

tesshuflower:
> Thank you @tesshuflower for the fix !

Thx to @JohnStrunk for figuring this one out!

shubham-pampattiwar:
> > Thank you @tesshuflower for the fix !
> 
> Thx to @JohnStrunk for figuring this one out!

Thank you @JohnStrunk Would love to hear about how you figured out the root cause !

tesshuflower:
Will try to explain as I understand it - When doing a sync in Direct/None mode with an RWO PVC, we need to make sure we schedule the mover pod on the same node as your running pod which is currently using the PVC.

To do this we find the node on which that pod is running and previously we set the `NodeName` on the mover job spec. However it seems that this doesn't go through normal scheduling - the result being our cache PVC (since the storageclass is using WaitForFirstConsumer) was never scheduled along with the POD that's trying to use it.  It seems that using `NodeSelector` instead of NodeName will go through normal scheduling (which then sets the NodeName itself on the pod) and our cache PVC can go to bound state.

kaovilai:
This PR has a bug @tesshuflower 

We hit an edge case where this PR does not apply perfectly.

See [gist](https://gist.github.com/kaovilai/c75dc582ca76c9b2d3a2f0bf1a141643)

```
podToBackupPVC.spec.nodeName == Node.metadata.name != Node.metadata.labels[kubernetes.io/hostname]
ip-10-0-253-51.ec2.internal == ip-10-0-253-51.ec2.internal != ip-10-0-253-51
```

PR #374, opened 2022-08-15
Update helm chart and manifests to deploy operator accd to pss/restricted

### LINKED ISSUE #373: Modify the VolSync operator Deployment to comply with PSS/Restricted
The VolSync operator Deployment should drop permissions such that it complies w/ the PSS/Restricted profile.
Note: Due to OCP 4.10- compatibility, this will not include Seccomp/RuntimeDefault.

JohnStrunk:
While we had the intent to already conform to PSS/Restricted, there was an error in the helm template that was overriding our hard-coded pod security settings. This removes the hardcode from the template and puts the proper defaults into the values file.

The OCP namespace autolabeler for PSS wasn't picking up on our RBAC configuration to use a special SCC, and so the Namespace wasn't being labeled properly. This PR updates the RoleBinding object created by VolSync to include the Role's Namespace so that the autolabeler sees it correctly.

This PR also adds explicit Namespace labeling to many of the kuttl tests (`01-label-namespace.yaml` files) in order to get them to run on the latest OCP builds. This effectively disarms the PSS enforcement in the test namespaces. Since we're going to be removing the kuttl tests soon (PR #353 should be the last of the porting), I took this shortcut. The ansible tests don't need the labeling because the test workloads are compliant.

**Related issues:**
Fixes #373


JohnStrunk:
/unhold
/cc @tesshuflower 

Finally :green_circle: 
This PR strayed from the intended purpose into "a just get s**t working" sort of PR, but the individual commits are reasonable (I hope).

tesshuflower (approved):
/lgtm

PR #377, opened 2022-08-18
helm: Make chart manage VolSync CRDs

### LINKED ISSUE #261: Better manage CRDs in Helm chart
Currently, we provide the VolSync CRDs in the `helm/crds` directory. CRDs provided here are automatically applied by Helm when it installs a chart, if the CRD does not exist in the cluster. If the CRD exists, the provided files are ignored.
[Helm's policy re CRDs](https://helm.sh/docs/topics/charts/#limitations-on-crds)

Between 0.3 and 0.4, we added a couple of fields to the CRD, but a user that is upgrading from 0.3 will not get the new CRD versions.
We need to restructure the chart to get around this limitation.

CRDs will continue to change as the operator evolves, and the installed version must be kept in sync with what the operator expects to see. The alternative to automatic management is requiring the user to manually apply the new/updated CRDs from the appropriate chart version. This is a poor user experience.

- New installs will have the associated CRD installed automatically
- Chart upgrades will apply any CRD changes to bring the CRDs up to date
- Chart uninstall will not remove the CRDs from the cluster

**Additional context**
CRDs can be installed (and templated) just like any other manifest as long as the rest of the Chart doesn't attempt to use the newly installed CRD (and we don't). So here's a *potential solution that needs to be validated*:
- Move the CRDs to be regular templates
- Wrap the declarations so that they are not removed on uninstall (`helm uninstall` doesn't remove CRDs)
  - I think this can be done by checking `Release.IsUpgrade` and `Release.IsInstall`

JohnStrunk:
Helm will only install CRDs if they are not present. It will not upgrade
or remove them. This change switches the CRDs to be managed as a part of
the templating mechanism so that they can be upgraded (and removed if
necessary).

This is a breaking change for Helm users as noted in the Changelog. In order for a chart upgrade to succeed, users will need to label/annotate the CRDs

**Related issues:**
Fixes #261

### DIFF
--- a/helm/volsync/README.md
+++ b/helm/volsync/README.md
@@ -86,6 +86,8 @@ Please see https://volsync.readthedocs.org for documentation.
 The following parameters in the chart can be configured, either by using `--set`
 on the command line or via a custom `values.yaml` file.
 
+- `manageCRDs`: true
+  - Whether the chart should install/upgrade/remove the VolSync CRDs
 - `replicaCount`: `1`
   - The number of replicas of the operator to run. Only one is active at a time,
     controlled via leader election.
--- a/helm/volsync/README.md
+++ b/helm/volsync/README.md
@@ -87,7 +87,7 @@ The following parameters in the chart can be configured, either by using `--set`
 on the command line or via a custom `values.yaml` file.
 
 - `manageCRDs`: true
-  - Whether the chart should install/upgrade/remove the VolSync CRDs
+  - Whether the chart should install/upgrade the VolSync CRDs
 - `replicaCount`: `1`
   - The number of replicas of the operator to run. Only one is active at a time,
     controlled via leader election.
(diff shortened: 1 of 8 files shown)


JohnStrunk:
/unhold
/cc @tesshuflower 
The CRD is now left in the cluster on uninstall

helm/volsync/README.md
```diff
@@ -86,6 +86,8 @@ Please see https://volsync.readthedocs.org for documentation.
 The following parameters in the chart can be configured, either by using `--set`
 on the command line or via a custom `values.yaml` file.
 
+- `manageCRDs`: true
+  - Whether the chart should install/upgrade/remove the VolSync CRDs
```
tesshuflower, line 90: should this comment not mention "remove" for the VolSync CRDs?

JohnStrunk: Good catch. fixed

tesshuflower (approved):
/lgtm

Issue #365, opened 2022-08-11
Support a Namespace annotation indicating elevated privileges are ok for movers


JohnStrunk:
**Describe the feature you'd like to have.**
The VolSync controller should look for a well-known annotation on the user's Namespace that would serve as an indicator that movers should be run with elevated privileges in that Namespace.

**What is the value to the end user? (why is it a priority?)**
Reducing the (default) permissions that are granted to movers enhances the security of the cluster by ensuring that movers (which run in the user's Namespace) only run w/ permissions that are equivalent to that of a normal user. Unfortunately, that will limit the ability of movers to preserve file metadata such as uid/gid and some attrs. It may also prevent reading (and replicating) some files. For this reason, it must be possible to securely signal that movers should be run w/ elevated privileges when these additional items need to be replicated.

By using an annotation on the Namespace, we ensure that an entity w/ elevated privileges has "ok-ed" the mover elevation.

**How will we know we have a good solution? (acceptance criteria)**
- Document the annotation that VolSync will use and explain the implications of the enhanced privileges: more accurate replication, but potential for user privilege escalation.
- The controllers should look for the annotation and make that information available as a flag via the mover interface (probably via the Builder)

**Additional context**
This item is just to provide the annotation to the interface. Changes to the behavior of movers will be handled separately.

JohnStrunk:
Proposed Namespace annotation: `volsync.backube/privileged-movers: true`

PR #387, opened 2022-08-22
Check for mover privilege annotation on ns and pass to movers

tesshuflower:

- Looks for annotation `volsync.backube/privileged-movers: true` on the namespace the ReplicationSource/Destination is running in to determine if the user is allowing privileged movers

- I put the annotation constant itself in api/v1alpha1 - in case any users want to access it - is there a better location for it?
- Currently this change is only looking at the namespace annotation - Anything else we should be looking for such as the kube version or the namespace labeling itself? 

**Related issues:**
Fixes https://github.com/backube/volsync/issues/365

### DIFF
--- a/controllers/replicationdestination_controller.go
+++ b/controllers/replicationdestination_controller.go
@@ -72,6 +72,7 @@ var _ sm.ReplicationMachine = &rdMachine{}
 //+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=serviceaccounts,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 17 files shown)


controllers/replicationdestination_controller.go
```diff
@@ -72,6 +72,7 @@ var _ sm.ReplicationMachine = &rdMachine{}
 //+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=serviceaccounts,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch
```
JohnStrunk, line 75: These permissions need to be added to the helm chart templates also

JohnStrunk (commented):
This looks good. I think all that's missing is the Helm chart update to add NS permissions

PR #388, opened 2022-08-22
restic: Dump the output of repo probe when we get an unexpected response

JohnStrunk:
- Trying to debug openshift/release#31509
- We should have this output anyway

**Related issues:**


tesshuflower (approved):
/lgtm

PR #397, opened 2022-08-26
syncthing: Disable use of minio's sha256-simd library

### LINKED ISSUE #394: Remove sha256-simd from Syncthing
In #391, we removed sha256-simd from restic due to fips concerns. The library is also imported by syncthing and should be removed from there as well.

This is *probably* needed for running on fips clusters.

**Additional context**

JohnStrunk:
This ensures we use the golang internal implementation so that we're compatible with the fips-enabled golang compiler.

Syncthing thankfully has built-in support for switching implementations, gated by an environment variable.

**Related issues:**
Fixes #394

### DIFF
--- a/mover-syncthing/entry.sh
+++ b/mover-syncthing/entry.sh
@@ -129,6 +129,11 @@ for op in "$@"; do
       # ensure our environment is configured before syncthing runs
       preflight_check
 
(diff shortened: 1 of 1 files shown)


tesshuflower (approved):
/lgtm

Nice that syncthing lets you override this

Issue #402, opened 2022-08-29
VolSync rsync mover source does not connect to destination if using an ipv6 address


tesshuflower:
**Describe the bug**

If a user creates a replication destination with ClusterIP in an ipv6 env and the address is an ipv6 address - if they use this address in their replicationsource, the replicationsource will not be able to connect to the destination properly.

Workaround is to use the address of the service created by the replication destination instead of the ipv6 address.

**Steps to reproduce**

- Create a replicationdestination in an ipv6 environment using `serviceType: ClusterIP` 
- The replicationdestination creates a service which has a clusterIP defined in the spec, e.g.

```
spec:
  clusterIP: fd02::3612
  clusterIPs:
  - fd02::3612
  ```

- If this address is used as the address in the replicationsource spec, it will fail to connect.

Workaround is to use the replicationsource service name instead of the ipv6 address.  e.g.

```
 address:  volsync-rsync-dst-database-destination.dest
 ```

**Expected behavior**

Expect the rsync mover to be able to connect to the destination using an ipv6 address.

This may be possible by modifying the mover to connect using:

```
rsync@[ipv6address]
```

Note that we cannot simpy direct a user to use something like [fd02::3612] as the address in the replicationsource as this causes the certs to fail checking the remote address which does not match.

**Actual results**

rsync mover pod on the replicationsource side has logs like this:

```
VolSync rsync container version: ACM-0.4.1-e6dde1b
Syncing data to fd02::3612:22 ...
rsync: getaddrinfo: fd02 873: Name or service not known
rsync error: error in socket IO (code 10) at clientserver.c(127) [sender=3.1.3]
Syncronization failed. Retrying in 2 seconds. Retry 1/5.
rsync: getaddrinfo: fd02 873: Name or service not known
rsync error: error in socket IO (code 10) at clientserver.c(127) [sender=3.1.3]
Syncronization failed. Retrying in 4 seconds. Retry 2/5.
rsync: getaddrinfo: fd02 873: Name or service not known
rsync error: error in socket IO (code 10) at clientserver.c(127) [sender=3.1.3]
Syncronization failed. Retrying in 8 seconds. Retry 3/5.
```

**Additional context**

PR #405, opened 2022-08-30
detect ipv6 dest addr and pass with [] to rsync

tesshuflower:

Rsync if provided with an IPv6 address needs to use the address in the format [ipv6addr].  This is a URL format.

In the mover script, attempt to detect if the destination address is an ipv6 address and wrap with [] only if it does not already contain [].  This hopefully avoids issues with urls that are already formatted correctly.

- Crude ipv6 check - checking that there are enough stanzas with : (or :: is found)
- Also avoids wrapping with [] if the address already contains [] - potentially could happen with a url address that embeds an ipv6 address?

**Related issues:**
https://github.com/backube/volsync/issues/402

### DIFF
--- a/mover-rsync/source.sh
+++ b/mover-rsync/source.sh
@@ -40,9 +40,18 @@ Host *
 SSHCONFIG
 
 URL_DESTINATION_ADDRESS=$DESTINATION_ADDRESS
-if [[ "$DESTINATION_ADDRESS" == *"::"* || "$DESTINATION_ADDRESS" == *":"*":"*":"*":"*":"*":"*":"* ]]; then
+
+# If we get a bare ipv6 address it must be wrapped with [] for rsync
+# Looking for either:
+# 1) 8 groups of hex digits separated by ":"
+# 2) a "::" in the string
+IPV6_REGEX='(([0-9a-fA-F]{0,4}:){7}[0-9a-fA-F]{0,4})|(::)'
+
+if [[ "$DESTINATION_ADDRESS" =~ $IPV6_REGEX ]]; then
   echo "Destination address $DESTINATION_ADDRESS is ipv6"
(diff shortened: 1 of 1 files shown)


mover-rsync/source.sh
```diff
@@ -39,19 +39,27 @@ Host *
   TCPKeepAlive no
 SSHCONFIG
 
+URL_DESTINATION_ADDRESS=$DESTINATION_ADDRESS
+if [[ "$DESTINATION_ADDRESS" == *"::"* || "$DESTINATION_ADDRESS" == *":"*":"*":"*":"*":"*":"*":"* ]]; then
+  echo "Destination address $DESTINATION_ADDRESS is ipv6"
+  if [[ "$DESTINATION_ADDRESS" != *"["*"]"* ]]; then
+    URL_DESTINATION_ADDRESS="[$DESTINATION_ADDRESS]"
+  fi
+fi
```
JohnStrunk, line 48: The matching makes me nervous. What about bash regular expressions:
```bash
# Looking for either:
# 1) 8 groups of hex digits separated by ":"
# 2) a "::" in the string
IPV6_REGEX='(([0-9a-fA-F]{0,4}:){7}[0-9a-fA-F]{0,4})|(::)'

if [[ "$DESTINATION_ADDRESS" =~ $IPV6_REGEX ]]; then
    echo "It's IPv6"
fi

if [[ ! "$DESTINATION_ADDRESS" =~ \[.*\] ]]; then
    echo "no brackets"
fi
```

tesshuflower: That's much nicer than what I had - I'm terrible with regex's so threw my version together fast before I lost that ipv6 env...I've done a bunch of different address checks and I think your suggestion looks good to me - will update...

Issue #404, opened 2022-08-30
rsync when using ROX access modes for the replicationsource - the source job fails to start


tesshuflower:
**Describe the bug**

When working with RamenDR where they are using Cephfs, a workaround is required to create a read-only pvc from snapshot for the replication source.  This avoids a performance issue with creating the pvc from snapshot.

So in the replicationsource, accessModes of [ ReadOnlyMany ] is specified so that volsync will create a cephfs ROX pvc from a snapshot.

However when using this PVC as the data source for the rsync replicationsource, we see a `CreateContainerError`

From the pod:

```
    started: false
    state:
      waiting:
        message: 'relabel failed /var/lib/kubelet/pods/c74f09dc-1953-49f5-a401-0f09c27c7f40/volumes/kubernetes.io~csi/pvc-e8a5d133-6149-41ce-b9d2-2aeb4b2d24a3/mount: lsetxattr /var/lib/kubelet/pods/c74f09dc-1953-49f5-a401-0f09c27c7f40/volumes/kubernetes.io~csi/pvc-e8a5d133-6149-41ce-b9d2-2aeb4b2d24a3/mount: read-only file system'
        reason: CreateContainerError
```

**Steps to reproduce**

Use ROX as the source volume in an rsync replicationsource.

**Expected behavior**

Ideally we should not need write access on the replicationsource side.

**Actual results**

**Additional context**

tesshuflower:
Unfortunately it looks like using readOnly: true for volumes in general does not always work for replicationsource

Possible solution:
- If the source PVC only supports ROX, then for the replicationsource specify `readOnly: true` for the volume in the job spec.  This allows ROX pvcs to be mounted and read, while for other readable PVCs we can maintain the current behavior.

tesshuflower:
/close

Fixed by: https://github.com/backube/volsync/pull/406

PR #406, opened 2022-08-30
rsync src mover - mount PVC as readOnly for ROX

tesshuflower:

In cases where the source PVC accessModes only supports ReadOnlyMany (ROX), mount the pvc as read-only in the ReplicationSource mover job.

- Not doing this for other PVCs as I have encountered permission errors when trying to access the volume mounted as read-only.  So far testing with CephFS and ROX pvcs it does not have issues when mounting read-only.

- Attempts to avoid changing for most current scenarios by mounting read-only only for replicationSource (where we do not need to write to the PVC) and only when the source PVC access modes only allow ROX.

- This change right now is confined to rsync - we could potentially attempt to do a similar change for other movers as well?

**Related issues:**
https://github.com/backube/volsync/issues/404

### DIFF
--- a/controllers/mover/rsync/mover.go
+++ b/controllers/mover/rsync/mover.go
@@ -371,6 +371,8 @@ func (m *Mover) ensureJob(ctx context.Context, dataPVC *corev1.PersistentVolumeC
 
 		runAsUser := int64(0)
 
+		readOnlyVolume := false
+
 		containerEnv := []corev1.EnvVar{}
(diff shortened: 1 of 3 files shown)


tesshuflower:
/cc @JohnStrunk 

I think this one is pretty well tested now from the ramen side - What do you think?

JohnStrunk:
/lgtm
We should probably open issue(s) to apply this to restic and rclone also.

tesshuflower:
> #404

Created issues for restic and rclone:
https://github.com/backube/volsync/issues/412
https://github.com/backube/volsync/issues/411

PR #407, opened 2022-08-31
Allow passing in GOOGLE_APPLICATION_CREDENTIALS

### LINKED ISSUE #349: Enable support for Google object storage credential file to be mounted on replication source/destination pods
Support for mounting GCS credential json file

We are currently using volsync with [OADP Operator ](https://github.com/openshift/oadp-operator) to move CSI snapshots to object storage. We were able to integrate volsync restic datamover for AWS S3 and Azure blob storage but are blocked on Volsyn restic with GCS.

Can you have a way to mount the GCS credentials json via a volume mounted on replication source/destination pods, so that the file path can be passed to the restic env var `GOOGLE_APPLICATION_CREDENTIALS` for GCS to work with restic datamover ?

Thank you in advance !!

Also, Awesome controller guys !!

JohnStrunk:
restic expects GAC to hold the path to a json file w/ the actual credentials, but we didn't have a method to actually mount the json file into the container. This change makes it possible to supply the json file **in the restic Secret** under the GAC key name. VolSync then takes the value and mounts that in the container and sets up an appropriate env var to point to it.

The usage of the GAC env var is not what is contained in the restic docs (a filename), instead, it's the actual file's contents.

**Related issues:**
Fixes #349

### DIFF
--- a/controllers/mover/restic/mover.go
+++ b/controllers/mover/restic/mover.go
@@ -47,6 +47,8 @@ const (
 	resticCache          = "cache"
 	resticCAMountPath    = "/customCA"
 	resticCAFilename     = "ca.crt"
+	credentialDir        = "/credentials"
+	gcsCredentialFile    = "gcs.json"
 )
 
 // Mover is the reconciliation logic for the Restic-based data mover.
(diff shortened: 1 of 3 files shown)


tesshuflower (approved):
/lgtm

Issue #366, opened 2022-08-11
Rclone: enable reduced mover permissions


JohnStrunk:

JohnStrunk:
/close
I think this work is complete for rclone

Issue #363, opened 2022-08-10
Permission reduction


JohnStrunk:
This issue tracks the work to reduce the permissions required by VolSync.

### Goals

- By default, both VolSync and the movers should run under the [Pod Security Standards](https://kubernetes.io/docs/concepts/security/pod-security-standards/) "Restricted" policy
- It should be possible to securely enable VolSync to run movers with elevated privileges when necessary to preserve permissions/UID/GID during replication

### Tasks

- [ ] Modify the VolSync operator Deployment to comply with PSS/Restricted
- [ ] Provide a well-known annotation that can be added to the user's Namespace that indicates movers should run there with elevated permissions
- [ ] Create a new rsync-based mover that uses stunnel w/ PSK instead of ssh to protect the transfer.
- [ ] Deprecate the old rsync/ssh mover since we can't sufficiently reduce its permissions
- [ ] Rclone: enable reduced mover permissions
- [ ] Restic: enable reduced mover permissions
- [ ] Syncthing: enable reduced mover permissions
- [ ] CLI: Enhance CLI to support "migration" via rsync/stunnel

tesshuflower:
Moving to version 0.8 milestone to cover cli enhancement and deprecating the rsync/ssh mover.

PR #413, opened 2022-09-06
Add package to detect platform properties

JohnStrunk:
Adds the "platform" package that probes the cluster to determine properties of the cluster. This is necessary for #363 so that the movers can adjust their Pod manifests properly.

Right now, all it does is detect whether it's OpenShift and whether the "restricted-v2" SCC is available on the cluster. These checks parallel what we had to do w/ the e2e test manifests to support both kube and OCP.
Detection of OpenShift is done by the presence of any SCC objects, and restricted-v2 is by checking for it specifically.

- Adds RBAC to the operator so that it can read SCCs

**Related issues:**
- #363
- #366

### DIFF
--- a/controllers/platform/properties.go
+++ b/controllers/platform/properties.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2022 The VolSync authors.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published
+by the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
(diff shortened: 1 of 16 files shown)


controllers/platform/properties.go
```diff
@@ -0,0 +1,90 @@
...
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Properties contains properties about the environment in which we are running
+type Properties struct {
+	IsOpenShift        bool // True if we are running on OpenShift
+	HasSCCRestrictedV2 bool // True if the StorageContextConstraints "restricted-v2" exists
```
tesshuflower, line 34: SecurityContextConstraints?  Just a typo in the comment

tesshuflower, line 34: typo in comment - StorageContextConstraints should be SecurityContextConstraints?

tesshuflower (approved):
/lgtm

PR #424, opened 2022-09-07
rclone: Don't write permission file into data volume

JohnStrunk:
Instead of writing the "permissions.facl" file into the data volume and syncing, this mounts a tempdir and places the file there.
- Benefit: Doesn't modify the user's data volume
- Drawback: Requires 2 rclone calls (one for data, one for the permission file)

**Related issues:**
- #366
- #411


tesshuflower (approved):
/lgtm

PR #426, opened 2022-09-09
Rclone: Add ability to run as non-root

JohnStrunk:
Adds the ability to run the rclone mover with a non-root UID.

- API change: Adds `moverSecurityContext` field to the rclone portion of the ReplicationSource/ReplicationDestination CRDs  
  This allows the user to directly specify the PodSecurityContext that will be used by the mover. The expectation is that they should set the PSC to match what they use for their primary workload.

Changes for restic:
- It adds CAP_DAC_OVERRIDE, CAP_CHOWN, and CAP_FOWNER so that it can properly restore file metadata
- An EmptyDir was added to /tmp to allow readonly rootfs

Changes for rsync:
- It explicitly names the caps that are allowed so that we can include the "drop: ALL" requirement. This meant adding: CAP_SETUID and CAP_SETGID for sshd in addition to the file permission ones listed, above.
- Two EmptyDir volumes were added to allow us to move to a readonly rootfs
- sshd was moved to container port 8022 to avoid the cap for using low numbered ports. This is not visible externally since the Service still listens on :22 and remaps the port number.

The SCC had been updated to be as restricted as I could make it:
- readonly root container image
- Require dropping all caps except those explicitly allowed. All that are allowed are commented with why.
- Added default seccomp profile
- Disallow privilege escalation 

**Related issues:**
#366


JohnStrunk:
/hold
Depends on #429

JohnStrunk:
/cc @tesshuflower
I seem to remember we were holding PR this for some reason, but I don't remember why. The backports perhaps?

tesshuflower (approved):
/lgtm

PR #429, opened 2022-09-12
tests: enhance roles to run w/ and w/o uid 0

JohnStrunk:
Updates the test roles to allow them to run both as a non-root user and as uid 0

I split this out of #426 since it's just changing the tests.

I changed the pvc_affinity_pod role to use a Deployment instead of a bare Pod so that OCP's admission controllers would set the default contexts such that it could run.

**Related issues:**
#426


tesshuflower (approved):
/lgtm

PR #457, opened 2022-10-04
Cvp e2e test img (custom-scorecard-tests)

tesshuflower:
- Adds code to create a custom scorecard image that packages the tests in test-e2e
- Also packages some prereq setup scripts from hack (run-minio.sh, ensure-default-csi.sh)
- Will also generate a scorecard config.yaml that can be pushed to the midstream

- May still need to update config.yaml to use the proper label of the scorecard image (in scorecard config.yaml)  in release-* branches.

**Related issues:**


tesshuflower:
/cc @JohnStrunk 

Hopefully I've done the updates in the .github workflows ok - do you mind looking at those in particular?

.github/workflows/operator.yml
```diff
@@ -227,7 +248,7 @@ jobs:
 
   e2e:
     name: End-to-end
-    needs: [build-operator, build-rclone, build-restic, build-rsync, build-syncthing, kubectl-plugin]
+    needs: [build-operator, build-rclone, build-restic, build-rsync, build-syncthing, build-scorecard, kubectl-plugin]
```
JohnStrunk, line 251: `build-scorecard` should be moved down to `e2e-success.needs` since it's not actually used during the e2e tests.

custom-scorecard-tests/go.mod
```diff
@@ -0,0 +1,41 @@
+module main
```
JohnStrunk, line 1: Please add a section to `.github/dependabot.yml` to update these deps. Should be something like:

```yaml
  - package-ecosystem: "gomod"
    directory: "/custom-scorecard-tests"
    schedule:
      interval: "weekly"
```

We should also put a check in the gh actions to ensure [go.mod and go.sum are in-sync](https://github.com/backube/volsync/blob/cac2570b518757f3b04076d49b7f4ce072cfa920/.github/workflows/operator.yml#L59-L68).

tesshuflower:
@JohnStrunk do you mind taking another look?

PR #478, opened 2022-10-26
Golang1.18 for release-0.4

tesshuflower:
Golang 1.17 is going out of service - we are planning on moving downstream builds to use a go 1.18 builder, but want to make the changes here too so it's consistent.


JohnStrunk (commented):
/lgtm

PR #479, opened 2022-10-26
CI: Rename sonarcloud config

JohnStrunk:
We're using the "automatic" mode of sonarcloud, and it looks like we have been using the wrong name for the config file. Hopefully this will fix the issue where upstream dependencies are not being ignored.

**Related issues:**


JohnStrunk:
/cc @tesshuflower 
I'm hoping this solves our problem w/ SonarCloud

tesshuflower (approved):
/lgtm

PR #480, opened 2022-10-26
Reduced mover permissions for restic

### LINKED ISSUE #367: Restic: enable reduced mover permissions
It should be possible to run the restic mover both with its current elevated permissions as well as with normal user permissions

Running as a normal user is sufficient for typical replication scenarios, and it improves the security of the cluster by not running Pods w/ elevated permissions in the user's Namespace.

- Mover uses the flag provided by #365 to determine how the mover pod should be invoked
- Mover container image is able to run as both a normal user (UID != 0) and as UID==0
- "Elevated permission" mode preserves permissions/uid/gid, etc. to the extent possible given restic
- "Normal permission" provides best-effort preservation of metadata

**Additional context**
- Depends on #365 
- Special attention to the restic cache volume may be necessary. In the case of OpenShift, the PVC will be automatically made accessible to the mover pod, but in vanilla kube, explicit setting of `fsGroup` or `supplementalGroups` may be necessary.

JohnStrunk:
Runs restic as a normal user (whatever is default for the namespace/environment) by default.

**Related issues:**
Fixes #367


tesshuflower (commented):
since there's a new e2e test, could you run 
```
make custom-scorecard-tests-generate-config
```

And check in the generated custom-scorecard-tests/config.yaml?  This will just ensure the config.yaml (that would need to be copied to the cicd midstream) is updated with the latest e2e tests.

I don't think this part needs to hold up this PR (and do a whole rebuild cycle) so I can do this separately after this merges if that makes it easier.  Wondering if I should put a check for this in the github workflows, and potentially run this make target as a prereq for other targets - maybe the `test-e2e` target?

tesshuflower (approved):
/lgtm

JohnStrunk:
> Wondering if I should put a check for this in the github workflows

I opened #481 for a solution.

PR #483, opened 2022-10-28
CI: Don't fail if we can't upload coverage results

JohnStrunk:
We've been getting a lot of CI failures because the codecov server has been unable to get information from the GitHub API. It seems to be a known problem w/ rate limiting. Since it's been happening for an extended period, I'd like to not fail builds when coverage can't be uploaded.

This will mean that coverage stats may be out of date (or we don't get the PR comment), but it doesn't seem good to hold up a PR just for this.

**Related issues:**

### DIFF
--- a/.github/workflows/operator.yml
+++ b/.github/workflows/operator.yml
@@ -149,7 +149,10 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           file: ./cover.out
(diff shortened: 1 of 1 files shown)


JohnStrunk:
/cc @tesshuflower
Refreshing this PR as codecov seems to be unreliable again.

tesshuflower (approved):
/lgtm

PR #494, opened 2022-11-03
e2e: Don't compare uid for rclone w/ priv mover

JohnStrunk:
When running tests for the privileged rclone mover, the uid of the files may not match due to how UIDs are handled on OpenShift. This skips verifying that uids match between src and dst volumes.

**Related issues:**

### DIFF
--- a/test-e2e/run_tests_in_parallel.sh
+++ b/test-e2e/run_tests_in_parallel.sh
@@ -20,17 +20,22 @@ TESTRC=$?
 if [[ $TESTRC == 0 ]]; then
     echo; echo "Tests completed successfully"
 else
+    FAILURES=""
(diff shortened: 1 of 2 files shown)


tesshuflower:
@JohnStrunk this is a bit confusing to me - I would've thought we'd hit something like this when using unprivileged as perhaps the uid of the dest mover pod might not match the source - but with privileged shouldn't the uids match?

JohnStrunk:
I'm still investigating to make sure my theory is correct... But I think we're having problems because uid assignment is inconsistent, and OpenShift is doing a chown on mount, causing the uid mismatch.

tesshuflower (approved):
/lgtm

I like the extra logging of the FAILURES btw :)

PR #497, opened 2022-11-04
e2e: Fix Rclone privileged test

JohnStrunk:
The rclone tests (privileged and unprivileged) were using the same s3 path, causing a race. This was only noticeable for the priv test because that one was preserving the UID of the files.

Also:
- Dumps logs from the "helper tools" roles pods.
- Backs out ignoring the mismatches of UIDs (#494)

**Related issues:**
Reverts #494

### DIFF
--- a/test-e2e/roles/compare_pvc_data/templates/job.yml.j2
+++ b/test-e2e/roles/compare_pvc_data/templates/job.yml.j2
@@ -20,6 +20,8 @@ spec:
             - |
               set -e -o pipefail
 
+              id
+
               echo "Validating contents: PVC1 -> PVC2"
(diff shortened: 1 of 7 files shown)


JohnStrunk:
/hold
Need to remove the last commit before merging. It's just there for debugging.

JohnStrunk:
/unhold
/test all
/cc @tesshuflower

The failures we had been seeing were due to the 2 rclone tests using the came object bucket, not a code or OpenShift artifact.  It was a race.

tesshuflower (approved):
/lgtm

JohnStrunk:
/retest
seems to be a bad day for launching clusters

PR #503, opened 2022-11-07
Ensure volsync mover scc

tesshuflower:
- Moves reconciling the privileged mover scc volsync uses to main - will attempt do to this at startup.
- Will create the volsync mover scc if it doesn't exist (only on OpenShift) -- `volsync-privileged-mover` is used as the name by default, but could be overridden with the --scc-name flag.  However note that doing this would mean the clusterrole would need to be manually updated so volsync can use the scc.
- Will update the volsync mover scc if it already exists on the system, as long as the volsync ownership label is on the SCC ("app.kubernetes.io/created-by": "volsync") via a merge patch.  This should allow us to update the scc in the future if needed.

- I put the code to create/patch the scc in the platform package - using some of the existing functions there to check if the cluster is OpenShift.  This was very helpful for implementing this - but I'm not sure it's actually used anywhere else just yet - any issues with this?
- Helm - I attempted to remove the scc from the helm templates so we just have our 1 way of ensuring the scc.  Hopefully I did all this correctly.  This also meant removing the openshift.forceSCC parameter from the values.yaml.
- I've also renamed the mover scc from `volsync-mover` to `volsync-privileged-mover`.  Primarily for the helm scenario.  This way existing helm users can get the mover deployed with the proper volsync labels, and thus continue to get scc updates in the future if we change anything.  However this does mean we would leave behind the old `volsync-mover` scc.

**Related issues:**
https://github.com/operator-framework/operator-lifecycle-manager/issues/2847


JohnStrunk:
/lgtm
Nice! :sunglasses:

Issue #364, opened 2022-08-11
Create a new rsync-based mover that uses stunnel w/ PSK instead of ssh


JohnStrunk:
**Describe the feature you'd like to have.**
We need a new data mover that is similar to the existing rsync over ssh mover, except that it should use stunnel (TLS/PSK) as the transport instead of ssh.

**What is the value to the end user? (why is it a priority?)**
- The existing rsync/ssh mover requires sshd on the destination, and sshd requires additional capabilities that we'd like to remove, specifically, CAP_AUDIT_WRITE.

**How will we know we have a good solution? (acceptance criteria)**
The overall target is feature parity with current rsync/ssh
- Within and cross-cluster 1:1 replication
- Support for all copyMethods
- Option to auto-generate or use a provided key on both the source and destination

**Additional context**
The motivation for an additional mover instead of just changing the existing one is for ease of managing permissions, deprecation, and subsequent removal.

JohnStrunk:
@pranavgaikwad Are you still planning to work on this? If so, could you provide an update?
We're trying to get an idea of the remaining work related to permission reduction.

PR #510, opened 2022-11-18
Add new rsync-tls mover

JohnStrunk:
Adds a new rsync-based mover that uses stunnel instead of ssh for the transport.

**Related issues:**
#364


tesshuflower (approved):
/lgtm

Issue #514, opened 2022-11-28
CI: OpenShift detection in unit tests


JohnStrunk:
**Describe the bug**
We're getting fairly regular, though intermittent, failures of the following test in CI.

```
• Failure [0.007 seconds]
A cluster w/o StorageContextConstraints [It] should NOT be detected as OpenShift 
/home/runner/work/volsync/volsync/controllers/platform/properties_test.go:43

  Expected
      <bool>: true
  to be false

  /home/runner/work/volsync/volsync/controllers/platform/properties_test.go:46
```

**Steps to reproduce**

**Expected behavior**

**Actual results**

**Additional context**
This seems to be a case of system state leaking between tests.

tesshuflower:
/assign @tesshuflower 

I was experiencing this before and I guess my fixes to cleanup CRDs at the end of tests didn't fix it entirely.  I think perhaps one solution is to cleanup the CRD itself at the end of each test and also wait for it to be gone.

PR #515, opened 2022-11-28
Unit test fix - cleanup SCC crd in-between tests

tesshuflower:

Tries to reduce errors with unit tests where sccs could be left behind when the next test in the same process starts.

**Related issues:**
Fixes: https://github.com/backube/volsync/issues/514


JohnStrunk:
@tesshuflower After you fix the conflict, is this ready for review?

tesshuflower:
> @tesshuflower After you fix the conflict, is this ready for review?

yes, this should be ready to go - let me do that now

PR #524, opened 2022-12-01
Fix auto gen chck for release branches

tesshuflower:
- custom scorecard config generation was always using "latest" which doesn't actually match what we want in release branches - where we want to use the custom-scorecard-image tagged with "release-x.y". This attempts to use the correct tag on the custom scorecard image.

**Related issues:**

### DIFF
--- a/.github/workflows/operator.yml
+++ b/.github/workflows/operator.yml
@@ -56,7 +56,7 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
 
       - name: Ensure go module files are up-to-date (main branch)
-        if: github.ref_name == github.event.repository.default_branch
+        if: github.base_ref == github.event.repository.default_branch
         run: |
           echo "Generating custom-scorecard-config for latest"
           make custom-scorecard-tests-generate-config
@@ -69,7 +69,7 @@ jobs:
           fi
 
       - name: Ensure custom-scorecard-tests config.yaml is up-to-date (release branch)
-        if: github.ref_name != github.event.repository.default_branch &&
+        if: github.base_ref != github.event.repository.default_branch &&
             github.ref_type == 'branch'
         run: |
           CUST_IMG_TAG="${{ github.ref_name }}"
--- a/.github/workflows/operator.yml
+++ b/.github/workflows/operator.yml
@@ -55,9 +55,26 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
 
-      - name: Ensure custom-scorecard-tests config.yaml is up-to-date
+      - name: Ensure go module files are up-to-date (main branch)
+        if: github.ref_name == github.event.repository.default_branch
         run: |
+          echo "Generating custom-scorecard-config for latest"
           make custom-scorecard-tests-generate-config
+          diff=$(git diff --color --ignore-space-change -- custom-scorecard-tests/config.yaml)
+          if [ -n "$diff" ]; then
+            echo "$diff"
+            echo "***** custom-scorecard-tests/config.yaml is out-of-date *****"
+            echo "*****     run 'make custom-scorecard-tests-generate-config'      *****"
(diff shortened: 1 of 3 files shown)


tesshuflower:
/cc @JohnStrunk 

Do you mind taking a look at this one?  I messed up previously with the auto-gen file check in that it normally updates the custom-scorecard-tests/config.yaml with the scorecard image using the `latest` tag.  however on release branches it should be using the `release-x.y` tag.  I'm not sure if there's a good way to test these changes before merging?

JohnStrunk:
I'm not sure of a good way to test other than to create a dummy repo and try different scenarios.

I think this patch is missing some logic though. The workflow runs based on several different triggers:
- PRs to main or release branches
- Pushes to main or release branches
- Version tag pushes
- Periodic/manual rebuilds (workflow_dispatch) to branches or tags

Making sure I understand what we're going for (identical to the other images):
- main should use ":latest"
- release-x.y should use ":release-x.y"
- tags should convert vX.Y.Z to "X.Y.Z"

The [registry-push](https://github.com/backube/volsync/blob/main/.github/workflows/registry-push.yml) has some of this logic, but it doesn't run for PRs, meaning its logic is only ~1/2 the cases. :disappointed:

Looking at what you've got:
- I'm concerned about the `github.base_ref == github.event.repository.default_branch` comparison. I'm not sure whether base_ref is just the name or the whole path (i.e., like `github.ref_name` vs `github.ref`).
- I don't think it's covering the "push"/"workflow_dispatch" cases.

tesshuflower:
> I'm not sure of a good way to test other than to create a dummy repo and try different scenarios.
> 
> I think this patch is missing some logic though. The workflow runs based on several different triggers:
> 
> * PRs to main or release branches
> * Pushes to main or release branches
> * Version tag pushes
> * Periodic/manual rebuilds (workflow_dispatch) to branches or tags
> 
> Making sure I understand what we're going for (identical to the other images):
> 
> * main should use ":latest"
> * release-x.y should use ":release-x.y"
> * tags should convert vX.Y.Z to "X.Y.Z"
> 
> The [registry-push](https://github.com/backube/volsync/blob/main/.github/workflows/registry-push.yml) has some of this logic, but it doesn't run for PRs, meaning its logic is only ~1/2 the cases. 😞
> 
> Looking at what you've got:
> 
> * I'm concerned about the `github.base_ref == github.event.repository.default_branch` comparison. I'm not sure whether base_ref is just the name or the whole path (i.e., like `github.ref_name` vs `github.ref`).
> * I don't think it's covering the "push"/"workflow_dispatch" cases.

Makes sense John, I think I do need to fix the PR and push cases - I think maybe it's not required on a tag?  Wonder if I could skip the tag push scenario as in that case files aren't changeing.

tesshuflower:
@JohnStrunk I've done some testing in a dummy repo and made these changes, what do you think?

- only run if github.ref_type == "branch" - so don't bother running this check on tags
- for pr then github.base_ref is set, which should be something like `main` or `release-0.6`
- for push then use github.ref_name instead, which should also be `main` or `release-0.6`

JohnStrunk:
I think that's fine. The tags shouldn't actually need a check since the code would already be in a branch.
As I type this, I've just realized that the check doesn't actually need to be done for branches, just PRs since that's how the code gets there in the 1st place. It certainly won't hurt; it should just never fail.

I'll review now.

tesshuflower:
Makes sense about pushes too, thought maybe if something ever got force-pushed then we'd run the check?  But you're probably right that it's not strictly necessary.

@JohnStrunk  sorry one more question - do I need to cherry pick this into release-0.6 too so it runs on prs/pushes into release-0.6?

JohnStrunk:
> do I need to cherry pick this into release-0.6 too so it runs on prs/pushes into release-0.6?

Yes.

PR #532, opened 2022-12-06
e2e: Synchronously delete the Syncthing pods

JohnStrunk:
With the change to using multiple NS for the syncthing tests, we are deleting the ST pods one-by-one. This change switches to doing it synchronously to hopefully avoid an error where the 2nd delete was not finding the pod and failing.

**Related issues:**


tesshuflower (commented):
/lgtm

I like it! It removes that timing window I was worried about - nice that it can be done in 1 step here.

PR #552, opened 2022-12-12
tests: Add retries to scorecard build

JohnStrunk:
The custom scorecard build has a tendency to fail when installing dependenties over the network. This adds a retry script we use in other places to automatically retry a few times before failing.

**Related issues:**


tesshuflower (approved):
/lgtm

PR #558, opened 2022-12-13
Consolidate volsync into a single container image

### LINKED ISSUE #448: Unified container image
Right now, we build several containers that must all be used together:
- Volsync operator
- restic mover
- rclone mover
- rsync/ssh mover
- syncthing mover

All 5 of these final images are based on the same base (ubi9-minimal). We should consider building a single image that contains all the components.

- Users need to be careful to use the same release across all container images
- Users in offline environments must mirror them all separately
- With a single image, it is more likely that it would already be present on a node

- #558
	- All components reside in a single container image
	- Invoking a mover or the operator is just a matter of using a different entrypoint in the container
	- Helm chart should ensure the same image is used for all components
	- CSV Updated to only reference the single image
	- Update GH actions to only build/push single image
- openshift/release#35130
- #588
  - Clean up mover directories (old Dockerfiles, Makefiles)

**Additional context**
A quick test of putting everything into a single container indicates we'll end up w/ 1 image of about 718 MB. The bulk of this is due to the rsync/ssh mover (522 MB standalone). The current sum of the separate containers is 1213 MB.

JohnStrunk:
Puts all volsync code (controller + all movers) into a single container image

**Related issues:**
Fixes #448


JohnStrunk:
/cc @tesshuflower

This should be ready for review.
OpenShift CI's "helm" variant is going to fail because it's still building and using the old images, but the operator is creating mover pods assuming the new layout (mover files being in their own subdirectory)
openshift/release#35130 is going to fix up CI once this merges.

onedr0p:
Nice work here @JohnStrunk this should hopefully make maintenance easier moving forward.

I am curious of some design decisions, why did you choose UBI? and what benefits do you see having the movers be built "offline" instead of the binary being pulled in from an official upstream source?

JohnStrunk:
> I am curious of some design decisions, why did you choose UBI? and what benefits do you see having the movers be built "offline" instead of the binaries (restic, rclone, etc..) being pulled in from an official upstream source?

I work for Red Hat. I'm using UBI because that's the base image in the downstream VolSync product, so matching them up makes things a good bit easier (only have to maintain 1). I'm not a fan of the large number of packages that are there by default, but it's what I have available.
The choice to build restic, rclone, and syncthing are also due to product constraints. They don't have official rhel packages, so they need to be built from source. Complicating things are issues such as fips requirements for certain crypto libraries that mean we need to add patches.

tesshuflower (commented):
/lgtm

Great stuff John!  Just out of curiosity, is there a reason for keeping the old mover Dockerfiles around?

JohnStrunk:
/retest

> is there a reason for keeping the old mover Dockerfiles around?
No. I'll do a cleanup PR once we get the OCP CI moved over.

PR #559, opened 2022-12-13
e2e: Fix namespace role returning incorrect list of ns

JohnStrunk:
The namespace role was returning more than just the newly requested namespaces in the namespace list, causing test failures.

This reworks the namespace creation:
- Namespaces are now named `test-xxxx-n` where `xxxx` is a random integer that will be common for the namespace creation request, and `n` is an integer index for the namespace within the request: `0 ... num_namespaces-1`.
- The temporary variables used in the role are now prefixed to prevent leakage outside the role
- Determining the number to create no longer leaks the `num_namespaces` variable to the outside

**Related issues:**

### DIFF
--- a/test-e2e/roles/create_namespace/tasks/main.yml
+++ b/test-e2e/roles/create_namespace/tasks/main.yml
@@ -2,35 +2,34 @@
 
 - name: Get number of namespaces to create (default is 1)
   ansible.builtin.set_fact:
-    num_namespaces: "{{ num_namespaces | default(1) }}"
+    nsrole_num_namespaces: "{{ num_namespaces | default(1) }}"
 
-- name: Build list of namespaces
+- name: Create namespace prefix
   ansible.builtin.set_fact:
-    namespace_name_prefixes: "{{ namespace_name_prefixes | default([]) + ['test-%s-' | format(item)] }}"
-  loop: "{{ range(0, (num_namespaces | int)) | list }}"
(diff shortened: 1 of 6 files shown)


test-e2e/roles/create_namespace/tasks/main.yml
```diff
@@ -2,35 +2,34 @@
...
-- name: Save name of test Namespaces
-  ansible.builtin.set_fact:
-    namespaces: "{{ namespaces | default([]) + [item.result.metadata.name] }}"
-  with_items: "{{ test_namespaces.results }}"
-
-# Only set namespace var when 1 namespace is being used
 - name: Save name of first namespace (useful for tests that use 1 namespace)
   ansible.builtin.set_fact:
     namespace: "{{ namespaces[0] }}"
-  when: num_namespaces | int == 1
```
tesshuflower, line 33: I think I specifically tried to set the global `namespace` var for only when 1 namespace was created because when looping and passing `namespace` to a sub task via a `loop_var`, I would get warnings from ansible about re-using the `namespace`  var (something along the lines of "the variable is already in use", and that another name should be used to avoid collisions).so I wasn't confident that the sub task would get the updated namespace var rather than the global one.

Example here:

https://github.com/backube/volsync/blob/main/test-e2e/roles/test_syncthing_cluster_sync/tasks/main.yml#L16-L21

JohnStrunk: Ok. I put it back in.

tesshuflower (approved):
/lgtm

PR #588, opened 2023-01-10
Remove extra build files

### LINKED ISSUE #448: Unified container image

JohnStrunk:
Now that we are building a single container image, these additional Makefiles and Dockerfiles are no longer needed.
This also removes some yamls that were from initial PoC development and testing.

**Related issues:**
Resolves #448


JohnStrunk:
/retest
Opened the PR before the update to CI happened :cry:

tesshuflower (approved):
/lgtm

PR #590, opened 2023-01-10
ci: Update the version of csi hostpath driver

JohnStrunk:
This updates the version of the CSI hostpath driver to be the latest supported on each kube version

**Related issues:**


tesshuflower (approved):
/lgtm

Issue #538, opened 2022-12-07
Return status info from movers


JohnStrunk:
**Describe the feature you'd like to have.**
Today, there is relatively little information about what is happening during the sync process. We publish events for things like snapshots/clones, but once the Job starts, there is no available information until it completes. I'd like to be able to return status updates from the data movers so that the information can be exposed to end users without requiring them to `kubectl logs <volsync-mover-job>`.

**What is the value to the end user? (why is it a priority?)**
It would be good for the mover jobs to be able to expose status information so that users can more easily see what the current status of synchronization is. Is it syncing? Is it failing to connect?

**How will we know we have a good solution? (acceptance criteria)**
- Movers can send short "status updates" to the controller
- The controller can publish the status updates to the associated CR's status field (probably via the Synchronizing condition)

**Additional context**
Updates we might want to publish:
- (Un)successful connection to a remote (i.e., is the network working)
- Transfer rates and/or progress information (x% complete; MM:SS remaining)
- Amount of data transferred
- Relevant timestamps w/ the above

Potential methods for returning information:
- Specially formatted log lines that could be scraped by the controller
- http endpoint that could be read by the operator

tesshuflower:
Some info about current logging from movers

Rclone:

Example output from replication source mover (with no changes to rclone settings):
```
VolSync rclone container version: ACM-0.6.0-ce9a280
getfacl: Removing leading '/' from absolute path names
2022/12/14 20:41:10 DEBUG : Setting --config "/rclone-config/rclone.conf" from environment variable RCLONE_CONFIG="/rclone-config/rclone.conf"
2022/12/14 20:41:10 DEBUG : rclone: Version "v1.59.0-beta.." starting with parameters ["rclone" "sync" "--checksum" "--one-file-system" "--create-empty-src-dirs" "--progress" "--stats-one-line-date" "--stats" "20s" "--transfers" "10" "/data" "rclone-data-mover:rclone-test-0-zx42b" "--log-level" "DEBUG"]
2022/12/14 20:41:10 DEBUG : Creating backend with remote "/data"
2022/12/14 20:41:10 DEBUG : local: detected overridden config - adding "{vECL7}" suffix to name
2022/12/14 20:41:10 DEBUG : Using config file from "/rclone-config/rclone.conf"
2022/12/14 20:41:10 DEBUG : fs cache: renaming cache item "/data" to be canonical "local{vECL7}:/data"
[...]

PR #593, opened 2023-01-12
Return status info from movers

tesshuflower:
Adds status.latestMoverStatus which will attempt to capture parts of the latest mover job logs (applicable to rsync, rsync-tls, rclone, restic atm).

- Currently each mover has its own filter function to filter lines
   - These filter on a line-by-line basis - but we could pass the entire log as well if more complex filtering is required
- Current log filters are only selecting specific lines, not modifying the lines themselves - this could be improved if we want to change what we save to the log.
   - Filters themselves also can be tuned if we want to log less/more or capture specific outputs - examples with expected outputs can be see in the logfilter_test.go files for each mover.
   - On job failures (i.e. error scenario), the movers are currently saving the entire log.  Generally these are short, but not sure if there are scenarios where the error log could be long
- Using env vars right now to allow overrides of the max number of lines to tail from the job pod and a debug mode to allow saving the entire log - Is this ok?

Here's an example of status in a replicationsource (latestMoverStatus is new):
```
status:
  conditions:
  - lastTransitionTime: "2023-01-11T16:59:04Z"
    message: Waiting for manual trigger
    reason: WaitingForManual
    status: "False"
    type: Synchronizing
  lastManualSync: once
  lastSyncDuration: 1m22.766397s
  lastSyncTime: "2023-01-11T16:59:04Z"
  latestMoverStatus:
    logs: |-
      no parent snapshot found, will read all files
      Added to the repository: 929 B (691 B stored)
      processed 1 files, 5 B in 0:00
      snapshot 062534bc saved
      Restic completed in 8s
    result: Successful
  restic: {}
```

**Related issues:**
https://github.com/backube/volsync/issues/538


tesshuflower:
/cc @JohnStrunk 

Do you mind taking a look and see what you think?

JohnStrunk:
I'm ok with the approach of putting a few log lines into the status on success. The error path causes me concern though...
Right now, we're not bounding the about of data that we could try to shove into the status. Large objects can hurt etcd performance... There's also a maximum size.

At a minimum, we need to put a limit on the size. It would be great if we could do the same regex thing for failures, but I don't have a good way to figure out what the possible (important) error strings would be.

tesshuflower:
> I'm ok with the approach of putting a few log lines into the status on success. The error path causes me concern though... Right now, we're not bounding the about of data that we could try to shove into the status. Large objects can hurt etcd performance... There's also a maximum size.
> 
> At a minimum, we need to put a limit on the size. It would be great if we could do the same regex thing for failures, but I don't have a good way to figure out what the possible (important) error strings would be.

Agreed - I'll put in another field (with a reasonably small default) for the max size we'll log to the status - we could still combine this with regex filters for error scenarios if we find some that give us the relevant data.

@JohnStrunk Are you ok with these values getting set by env vars on the volsync operator?

PR #598, opened 2023-01-18
Remove replaces 0.5.1 from csv

tesshuflower:
- This is so we can publish to the ocp 4.12 catalog downstream the version being replaced needs to exist in the channel within the catalog, so we're removing it so we can "start again".
- Disadvantage is after publishing 0.5.2 it means users cannot bypass 0.5.2 and install 0.5.1 or 0.5.0 if they wanted.

**Related issues:**


JohnStrunk (commented):
/lgtm

PR #600, opened 2023-01-20
Annotate gh action versions w/ their tag

### LINKED ISSUE #492: Update GH actions w/ comments documenting current action version
We have set the SHA hash for all our GH actions according to best practice. Unfortunately, this obscures which version (vX.Y.Z) we are using. dependabot has recently added a feature where it will update a version number in a comment when it bumps the SHA.

We should add version comments so that dependabot will start updating them. Example:
https://github.com/dependabot/dependabot-core/blob/b4112ce4639d7eed1e3b2e0792eb7533f7cb125f/github_actions/spec/fixtures/workflow_files/pinned_sources_version_comments.yml#L7

**Additional context**
- dependabot/dependabot-core#5951

JohnStrunk:
Annotates all the GH actions hashes with their corresponding tag

**Related issues:**
Fixes #492

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -18,7 +18,7 @@ jobs:
 
     steps:
       - name: Checkout source
-        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c  # v3.3.0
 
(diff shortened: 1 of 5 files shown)


JohnStrunk:
/cc @tesshuflower 
This is just housekeeping... no code change. dependabot will update the comments when it bumps the versions.

tesshuflower (approved):
/lgtm

PR #604, opened 2023-01-23
rsync-tls: add logfilter for failed mover job

tesshuflower:

- rsync-tls logs a lot more detail than our other movers - when testing some scenarios where the destination could not be reached from the mover, the log saved ended up not having much useful error information as it would only catch the end of the log
- This is an attempt to catch some useful error strings

Update:
- Also default tail all logs from mover
- By default do not limit the lines we get from the mover pod
  when getting logs - these will still be filtered & possibly
  truncated before saving to the latestMoverStatus.
- Main motivation is rsync-tls is logging even more than I thought
  previously so log lines were being missed.

**Related issues:**
#538

### DIFF
--- a/controllers/mover/rsynctls/logfilter.go
+++ b/controllers/mover/rsynctls/logfilter.go
@@ -21,15 +21,34 @@ import (
 	"regexp"
 )
 
-var rsyncRegex = regexp.MustCompile(
+var rsyncTLSRegex = regexp.MustCompile(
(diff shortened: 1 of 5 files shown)


tesshuflower:
/cc @JohnStrunk 

Not sure what the deal is with the sonar failure - some sort of new check for the user in the Dockerfile?

JohnStrunk (commented):
/lgtm

PR #606, opened 2023-01-23
e2e: Try to make syncthing pod killing more robust

JohnStrunk:
The task to kill the running syncthing pods would sometimes fail even though they had successfully killed the pod. This changes the delete to just use the pod label directly. That required getting rid of the wait in the delete operation. To substitute, I added a 30s sleep after all the delete operations. This will give the controller a chance to re-poll all the instances and update its view of connection state (in the CR's status field) so that the following tasks get the proper information.

I'm hoping this address some of the CI failures we've been seeing, particularly in openshift CI.

**Related issues:**


tesshuflower (approved):
/lgtm

:thumbsup:

Issue #596, opened 2023-01-18
Pass annotations to Service objects


nesanton:
**Describe the feature you'd like to have.**
When defining the `ReplicationDestination` that defines a `serviceType`, I'd like to have an option of specifying annotations that would be passed to the Service object.
My use case: On Azure the services of type `LoadBalancer` can be annotated with `service.beta.kubernetes.io/azure-load-balancer-internal: "true"` which makes them private. By default such services are created with public (internet facing) scope.

**What is the value to the end user? (why is it a priority?)**
Right now there's no way to create internally scoped `LoadBalancer` services via the volsync operator. Unaware users may experience unexpected costs related to traffic passing out and in their cloud premise.

**How will we know we have a good solution? (acceptance criteria)**
Defining a `ReplicationDestination` for e.g. rsync type of replication with `serviceType: LoadBalancer` and a new additional field e.g. `serviceAnnotations: []` passes the annotations to the Service object. In case of Azure this results in creation of an internally scoped load balancer.

```yaml
spec:
  rsync:
    serviceType: LoadBalancer
    serviceAnnotations:
      - service.beta.kubernetes.io/azure-load-balancer-internal: "true"
```

**Additional context**
Kubernetes on Azure, OpenShift Container Platform on Azure, ARO

PR #607, opened 2023-01-26
svc annotations in spec (rsync & rsync-tls)

tesshuflower:
Fixes: https://github.com/backube/volsync/issues/596

- Adds a new field in the spec for rsync and rsyncTLS called "serviceAnnotations" to allow a user to set annotations on the service created by a replicationdestination.

Example usage:

```
apiVersion: volsync.backube/v1alpha1
kind: ReplicationDestination
metadata:
  name: rdest-tls-stuff
  namespace: ttest
spec:
  rsyncTLS:
    serviceType: ClusterIP
    serviceAnnotations:
      "a": "b"
      "service.beta.kubernetes.io/azure-load-balancer-internal: "true"
    copyMethod: Snapshot
    capacity: 2Gi
    accessModes: [ReadWriteOnce]
```

- I used a pointer to map in the types
  ```
  ServiceAnnotations *map[string]string `json:"serviceAnnotations,omitempty"`
  ```

  This is to allow a nil map (the default) which means VolSync will add its own default annotation(s).  But it also means if a 
  user specifies a map, those values will be used instead of any VolSync defaults.

  i.e. if you specify an empty map it means VolSync will not place any annotations on the service.

  For example:
  ```
  spec:
    rsyncTLS:
      serviceAnnotations: {}
  ```

- Additionally, I did not put serviceAnnotations in the replicationSource spec.  Rsync seems to allow the source to set the service up - but rsyncTLS does not - I was assuming this was intentional, so updated rsyncTLS to make sure the svc is not created on a replicationsource - is this correct?  Previously if the user did not specify address on the rsyncTLS replicationsource, it would go ahead and create a clusterIP service.

**Related issues:**
 https://github.com/backube/volsync/issues/596

### DIFF
--- a/api/v1alpha1/replicationdestination_types.go
+++ b/api/v1alpha1/replicationdestination_types.go
@@ -95,6 +95,11 @@ type ReplicationDestinationRsyncSpec struct {
 	// SSH connections.
 	//+optional
 	ServiceType *corev1.ServiceType `json:"serviceType,omitempty"`
+	// serviceAnnotations defines annotations that will be added to the
+	// service created for incoming SSH connections.  If set, these annotations
+	// will be used instead of any VolSync default values.
+	//+optional
+	ServiceAnnotations *map[string]string `json:"serviceAnnotations,omitempty"`
 	// address is the remote address to connect to for replication.
 	//+optional
 	Address *string `json:"address,omitempty"`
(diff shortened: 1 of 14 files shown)


JohnStrunk (commented):
/lgtm
Nice!

PR #612, opened 2023-02-01
Include cluster wide proxy env vars on restic mover

dymurray:

If a user installs volsync via OLM, the cluster-wide proxy settings are injected as env vars. These env vars need to be set on the mover pods so that restic can use a custom transport.

I need to add tests
**Related issues:**

https://github.com/backube/volsync/issues/613

https://issues.redhat.com/browse/ACM-3216

### DIFF
--- a/controllers/mover/syncthing/mover.go
+++ b/controllers/mover/syncthing/mover.go
@@ -444,29 +444,35 @@ func (m *Mover) ensureDeployment(ctx context.Context, dataPVC *corev1.Persistent
 		podSpec.ServiceAccountName = sa.Name
 		podSpec.RestartPolicy = corev1.RestartPolicyAlways
 		podSpec.TerminationGracePeriodSeconds = pointer.Int64(10)
+
+		envVars := []corev1.EnvVar{
+			{Name: configDirEnv, Value: configDirMountPath},
+			{Name: dataDirEnv, Value: dataDirMountPath},
+			// tell the mover image where to find the HTTPS certs
+			{Name: certDirEnv, Value: certDirMountPath},
+			{
+				Name: apiKeyEnv,
+				ValueFrom: &corev1.EnvVarSource{
+					SecretKeyRef: &corev1.SecretKeySelector{
+						LocalObjectReference: corev1.LocalObjectReference{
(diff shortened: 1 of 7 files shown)


tesshuflower:
/cc @JohnStrunk 

Do you mind reviewing John?  I think this should cover the immediate fix part of https://github.com/backube/volsync/issues/613

When it comes to cherry-picking into `release-0.6` I think that should wait until ACM 2.7/VolSync 0.6.0 ships next week, just to be safe

controllers/mover/syncthing/mover.go
```diff
@@ -490,6 +496,9 @@ func (m *Mover) ensureDeployment(ctx context.Context, dataPVC *corev1.Persistent
 			},
 		}
 
+		// Cluster-wide proxy settings
```
JohnStrunk, line 499: Isn't this redundant w/ L475?

tesshuflower: Good catch! That was my mistake  :grimacing:

JohnStrunk (commented):
/lgtm

JohnStrunk:
/approve
Forgot that one

controllers/utils/utils.go
```diff
@@ -114,3 +115,25 @@ func PvcIsReadOnly(pvc *corev1.PersistentVolumeClaim) bool {
 	// All other access modes support write
 	return false
 }
+
+func AppendEnvVarsForClusterWideProxy(envVars []corev1.EnvVar) []corev1.EnvVar {
+	httpProxy, ok := os.LookupEnv("HTTP_PROXY")
+	if ok {
+		envVars = append(envVars, corev1.EnvVar{Name: "HTTP_PROXY", Value: httpProxy})
+		envVars = append(envVars, corev1.EnvVar{Name: "http_proxy", Value: httpProxy})
+	}
```
kaovilai, line 124: @dymurray a bit late now.. but I would've preferred
```go
import(
...
	"github.com/operator-framework/operator-lib/proxy"

...

	envVar = append(envVar, proxy.ReadProxyVarsFromEnv()...)
```

PR #615, opened 2023-02-05
Update to latest restic and avoid vendored copies of restic and minio.

Yeicor:

- Updates restic to the latest version (v0.15.1).
- Deletes the vendored copies at mover-restic/restic and mover-restic/minio.
  - The required patch is performed at build time (/Dockerfile).
  - This makes it closer to the other tools and cleans up the repository.

Thanks to https://github.com/minio/minio-go/pull/1749, it is no longer needed to patch its source code to enable FIPS compatibility, adding the `fips` build tag should be enough. I've done this in the Dockerfile, but I didn't test FIPS compliance.

**Unrelated comments:**

- Thanks for maintaining this project!
- Which quay.io tag should be used for testing the newest features? `release-0.5`?


JohnStrunk:
> Which quay.io tag should be used for testing the newest features? release-0.5?

`release-0.6` will be the basis for the next tagged release.
`latest` tracks the active development in main

-----

Thanks for this PR. Glad to see that restic picked up that minio-go version.
I'd like to merge this change, but it has carry-on effects on our downstream build process. I'll follow up here once we get that sorted out.

JohnStrunk:
/approve
/cc @tesshuflower 

This seems ok from a downstream PoV, but will give Tesshu a chance to look further and lgtm.

Dockerfile
```diff
@@ -53,12 +53,20 @@ RUN make rclone
...
 
+RUN git clone --depth 1 -b ${RESTIC_VERSION} https://github.com/restic/restic.git
 WORKDIR /workspace/restic
 
-RUN go run build.go
+# Make sure the restic version tag matches the git hash we're expecting
+RUN /bin/bash -c "[[ $(git rev-list -n 1 HEAD) == ${RESTIC_GIT_HASH} ]]"
+
+# HACK: Apply our patch to restic to enable FIPS mode
+RUN find . -name '*.go' -exec sed -ri 's|github.com/minio/sha256-simd|crypto/sha256|' {} \; && go mod tidy
```
tesshuflower, line 66: Is this step still necessary?  Unfortunately downstream we are not able to edit source before building - so we'd need to keep vendoring if we want to do this.  I'm hoping the fips build tag will solve the minio-go issues for fips, but now I'm uncertain (I'm struggling to remember the details of all this) if sha256-simd is also used directly by restic for some operations.

tesshuflower (changes requested):

Dockerfile
Yeicor: Yes, restic still uses `sha256-simd` to compute hashes internally. 
The clean solution would be to open a pull request for fips compatibility in `restic` if a tag is given, similar to https://github.com/minio/minio-go/pull/1749. Then, no patching would be necessary.
I ended up using another backup solution for my project, so I won't continue working on this.

tesshuflower: /close

Based on all the discussion we're going to close this for the moment.

PR #623, opened 2023-02-16
Add RunAsAny security context to mover privileged SCC

dymurray:

Changes security context field in privileged mover SCC to `RunAsAny`.

**Related issues:**


tesshuflower:
After some discussion, we've decided on going ahead with this feature: https://github.com/backube/volsync/issues/633 instead of updating the SCC in this PR.

Which would mean OADP could create their own service account with access to privileged SCC and tell VolSync to use that for the mover job.

PR #641, opened 2023-02-23
Update changelog for prev 0.5.2 and also update changelog for upcoming 0.6.1 release

tesshuflower:

**Related issues:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - kube-rbac-proxy upgraded to 0.14.0
 - All movers, except rsync-ssh, now run with reduced privileges by default (see docs)
 
+## [0.6.1]
+
+### Fixed
+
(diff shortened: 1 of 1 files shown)


tesshuflower:
/cc @JohnStrunk 

I tried to update the changelog for 0.5.2 and also for the upcoming 0.6.1 - note that I tried to also update the version to 0.6.1 at first, but it caused an issue as the prow ci-index-volsync-bundle-gen step fails - when using 0.6.1 REPLACES 0.6.0.  The packaging fails as 0.6.0 doesn't already exist in the bundle it's creating (I think this is the issue anyway).  Looking back through the history perhaps we never hit this as we always versioned the .z releases in the release-x.y branch itself? 

If it's ok with you I'll try to keep this PR to just the changelog updates, cherry-pick into release-0.6 and then do a PR directly into release-0.6 with 0.6.1 version updates in version.mk, the csv and the helm Chart.yaml.  After which I can cherry-pick https://github.com/backube/volsync/pull/612 into release-0.6.

Issue #633, opened 2023-02-22
Allow a serviceAccount to be specified in the ReplicationSource/ReplicationDestination


tesshuflower:
**Describe the feature you'd like to have.**

Allow a service account to be specified (by name) in the ReplicationSource/Destination specs.

This would be for advanced users and would be the service account that will be used for the mover job.

- Users would need to setup permisisons/privileges on the service account properly for the mover they are using.  For example for movers running in privileged mode they would need to setup access to the SCC they need, etc.
- The service account for the mover would not be created if the service account is set in the spec
- It would be assumed that the service account should exist in the namespace of the replicationsource/destination and if not there, the mover job will not be able to run.

**What is the value to the end user? (why is it a priority?)**

- ODAP will use this for 2 reasons:
   - To allow OADP to use their own service account with access to the privileged SCC in order for them to be able to set seLinux advanced parameters - they have a specific scenario where a read-only volume that originated in another namespace will be used and they need to set permissions accordingly.
   - Also to reduce the number of service accounts when customers have many replicationsources (many PVCs).  This way they can share a service account they create rather than having volsync create one per replicationsource.

**How will we know we have a good solution? (acceptance criteria)**

**Additional context**

tesshuflower:
@dymurray @weshayutin @shubham-pampattiwar FYI

tesshuflower:
/close

This is implemented in v0.7.0

PR #645, opened 2023-02-23
add moverServiceAccount to spec for advanced usage

tesshuflower:
- Allows user to specify a service account by name to use for mover jobs/deploys.
- The service account must exist in the namespace and have the privileges required by the mover.

I put the moverServiceAccount field in the spec under each mover specific section (such as spec.restic.moverServiceAccount) - however this is sort of a global field - should it be directly under spec?  I tried to be consistent as we have other fields such as moverSecurityContext already in the mover section.

**Related issues:**
#633


tesshuflower:
/cc @JohnStrunk 

I wasn't sure specifically about putting the moverServiceAccount field from the spec into each mover specific section, I thought it was the most consistent when comparing to the other spec params - however we could put this straight into spec (spec.moverServiceAccount) if you think this makes more sense.

tesshuflower:
FYI @JohnStrunk

Sorry, I had originally added you for a review too early - I had to make a fix - was reconciling the svc acct for rsync as unprivileged when the annotation wasn't on the namespace (like other movers) but rsync needs to be privileged all the time.  This fixes the e2e tests.

I think it's ready for review now.

JohnStrunk:
/lgtm

I like how you put SAHandler behind an interface.

PR #672, opened 2023-03-10
fix nodename empty

big-appled:
* when pod is pending, there is a case that pod does not has nodename. So the nodeselector in volsync pod should be not set.

**Related issues:**

### DIFF
--- a/controllers/utils/affinity.go
+++ b/controllers/utils/affinity.go
@@ -93,6 +93,9 @@ func AffinityFromVolume(ctx context.Context, c client.Client, logger logr.Logger
 
 func getNodeSelectorForNode(ctx context.Context, c client.Client, logger logr.Logger,
 	nodeName string) (map[string]string, error) {
+	if nodeName == "" {
+		return nil, nil
+	}
 	node := &corev1.Node{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: nodeName,
--- a/controllers/utils/affinity_test.go
+++ b/controllers/utils/affinity_test.go
@@ -31,8 +31,8 @@ import (
 var _ = Describe("Volume affinity", func() {
 	logger := zap.New(zap.UseDevMode(true), zap.WriteTo(GinkgoWriter))
 	var ns *corev1.Namespace
-	var rwxPVC, rwoBoth, rwoPending, rwoNone, vsOnly *corev1.PersistentVolumeClaim
-	var runningPod, pendingPod, vsPod *corev1.Pod
+	var rwxPVC, rwoBoth, rwoPending, rwoPendingWithoutNodeName, rwoNone, vsOnly *corev1.PersistentVolumeClaim
+	var runningPod, pendingPod, pendingPodWithoutNodeName, vsPod *corev1.Pod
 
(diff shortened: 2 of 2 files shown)


tesshuflower:
Thanks @big-appled for your proposed fix.  Could you give us some background on this error scenario?  I think normally when using CopyMethod: Direct, we would want to copy the affinity settings (if needed) from the pod, however you're right we would get errors while the POD is in Pending since no node is set.  We're trying to avoid proceeding (and perhaps not setting the affinity correctly) if the POD will eventually come up - is there a case where you expect the POD to remain in Pending state?

big-appled:
> Thanks @big-appled for your proposed fix. Could you give us some background on this error scenario? I think normally when using CopyMethod: Direct, we would want to copy the affinity settings (if needed) from the pod, however you're right we would get errors while the POD is in Pending since no node is set. We're trying to avoid proceeding (and perhaps not setting the affinity correctly) if the POD will eventually come up - is there a case where you expect the POD to remain in Pending state?

I am syncing data from primary pod (which is running) to secondary pod (which is pending by design) for disaster recovery. In this case, pvc is not actually mounted to the pod. When using direct mode, I expect the staging pod running to mount the pvc.

tesshuflower:
> > Thanks @big-appled for your proposed fix. Could you give us some background on this error scenario? I think normally when using CopyMethod: Direct, we would want to copy the affinity settings (if needed) from the pod, however you're right we would get errors while the POD is in Pending since no node is set. We're trying to avoid proceeding (and perhaps not setting the affinity correctly) if the POD will eventually come up - is there a case where you expect the POD to remain in Pending state?
> 
> I am syncing data from primary pod (which is running) to secondary pod (which is pending by design) for disaster recovery. In this case, pvc is not actually mounted to the pod. When using direct mode, I expect the staging pod running to mount the pvc.

This is a tricky one - The scenario we're worried about is with an RWO PVC in direct mode that needs to be scheduled on the same node as the users' pod.  When that pod is in pending, we has assumed we'd rather error and keep trying as if we go ahead and start the mover pod with no nodeselector, we will get scheduled automatically and perhaps prevent the users' pod from being able to access the pvc or not be able to access the pvc from our mover job.  @JohnStrunk do you have any thoughts on this?

JohnStrunk:
Adding a bit more context to this...

We need to watch out for interactions with storage topology. For StorageClasses using WaitForFirstConsumer, the first time a volume gets used determines which storage zone it will be provisioned into, forever locking it to that zone. Examples include EBS (where it gets locked to an AZ) or node-local storage like LVM-based PVs.

If the workload pod isn't assigned to a node and we let the mover pod start, it will cause the temporary volumes (like the restic cache or Syncthing config volume) to get provisioned, potentially in an incompatible zone from the primary workload pod that may have restrictions for various reasons (storage or otherwise). Once that happens, manual intervention would be required to fix it.

If there's no workload pod running, there's not much we can do... But, if there is a workload pod, we intentionally wait for it. We don't know why it's pending. It could be intentional, like in your case, but it's normally because there is some other unsatisfied condition. If we let the mover start, it may make the situation worse.

PR #675, opened 2023-03-14
Make v0.5.3 version

tesshuflower:

Due to issues building downstream on OCP 4.12 the downstream builds needed to be split into 2 separate operator bundle builds, one for ocp <= 4.11 and ocp >= 4.12.

No code changes planned for v0.5.3, but making a new version so the downstream operator can be built again with 1 consistent bundle build for all OCP versions, and then freshmaker rebuilds with base img updates can also resume.

**Related issues:**


birsanv (approved):
lgtm

PR #688, opened 2023-03-22
v0.7.1 for upcoming .z release

tesshuflower:

Update version metadata for upcoming v0.7.1 release - will need to be cherry-picked into release-0.7

**Related issues:**


tesshuflower:
/close

Closing due to errors with the ci/prow/ci-index-volsync-bundle automation not liking the replaces rules - this is not an issue with .Y updates as we publish to new channels and therefore don't use replaces in the CSV.  Will instead make 0.7.1 changes directly against release-0.7.  Will version main to 0.8.0 first just so release-0.7 doesn't appear ahead of main.

Issue #678, opened 2023-03-17
VolSync Metrics Service - needs unique pod selector


tesshuflower:
**Describe the bug**

The VolSync Metrics service uses a common pod selector:

```
 selector:
    control-plane: controller-manager
```

This is a common name generated by operator-sdk and can lead to exposing metrics from other operator pods in the namespace.

**Steps to reproduce**

**Expected behavior**

**Actual results**

**Additional context**

tesshuflower:
Targeting this to be fixed in v0.7.1

tesshuflower:
Label selectors have been updated to include more than just the `control-plane` label.

Here's the updated ServiceMonitor example that can be used:

https://github.com/backube/volsync/blob/8af0bf09979b8a9ceb75f0a1ca47c7cc53f2a868/config/prometheus/monitor.yaml

PR #696, opened 2023-03-23
volsync metrics svc in operator use unique label selector

tesshuflower:
Fixes: https://github.com/backube/volsync/issues/678

Updates the pod selector in the volsync metrics service to be more unique by also checking for the label:

```
 app.kubernetes.io/part-of: volsync
```

Also updates the deploy to use this label for its pods, and the ServiceMonitor example config/prometheus/monitor.yaml

Originally had planned to simply update the config/default/kustomization.yaml to set `commonLabels` which would automatically put the additional label on the deployment pods as well as the service (and their selectors).

However it turns out this breaks upgrades, as spec.selector is an immutable field on deployments, and when the spec.selector field is updated in the deployment within the CSV file, OLM gets into a failure loop trying and failing to update the volsync controller-manager deployment.

So the solution now adds labels on deployment pods and updates the service selector which is updatable and operator upgrades work.

**Related issues:**

https://github.com/backube/volsync/issues/678

### DIFF
--- a/bundle/manifests/volsync-controller-manager-metrics-service_v1_service.yaml
+++ b/bundle/manifests/volsync-controller-manager-metrics-service_v1_service.yaml
@@ -18,6 +18,7 @@ spec:
     protocol: TCP
     targetPort: https
   selector:
+    app.kubernetes.io/part-of: volsync
     control-plane: controller-manager
 status:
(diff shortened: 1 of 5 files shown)


birsanv (approved):
lgtm

PR #703, opened 2023-03-30
Updates for v0.7.1 release

tesshuflower:

This is to update the CHANGELOG and helm chart.yaml (maybe I should cherry-pick into release-0.7 just to be consistent too).

We ended up having to ship the downstream 0.7.1 operator so I'm going to go ahead and do the 0.7.1 version in VolSync as well.

The operator needed to be shipped as we almost had that same downstream issue again where ocp 4.14 wasn't added to the support matrix - luckily the downstream staging catalog hasn't been seeded for 4.14 yet, so this means we need to get another v0.7.Z out there with ocp 4.14 in the matrix before this happens.

**Related issues:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [unreleased]
 
+## [0.7.1]
+
 ### Changed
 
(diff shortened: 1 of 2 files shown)


tesshuflower:
/cc @JohnStrunk 

fyi @birsanv

Issue #673, opened 2023-03-10
custom CA w/ restic mover in a ConfigMap


seb54000:
**Describe the feature you'd like to have.**
There is an existing feature to specify a custom CA when contacting remote restic endpoint : https://github.com/backube/volsync/pull/398
The feature actually take a secret ressoruce to hld the CA certificate. We would like to be able to hold the custom CA in a ConfigMap

**What is the value to the end user? (why is it a priority?)**
Our useCase is that we use a functionality in openshift (that may exists in Kubernetes upstream, not sure) that creates a configMap automagically with the custom CA the admin deployed in the cluster, this is don through an annotation on an empty configMap
```
  labels:
    config.openshift.io/inject-trusted-cabundle: 'true'
```
https://docs.openshift.com/container-platform/4.9/networking/configuring-a-custom-pki.html#certificate-injection-using-operators_configuring-a-custom-pki 

**How will we know we have a good solution? (acceptance criteria)**
Possibility to specify either a secret or a configMap maybe. Should be exclusive
```
    customCA:
      key: ca-bundle.crt
      secretName: ca-secret

or 
    customCA:
      key: ca-bundle.crt
      configmapName: ca-cm
```

**Additional context**
Related initial issue : https://github.com/backube/volsync/issues/348

seb54000:
Thanks a lot for this improvement, do you know next date of an official release and integration in the operatorHub of this version ? For production cluster, we only go through released version.

If I'm not wrong last release is 0.7.1 and does not include this improvement

PR #706, opened 2023-04-03
Allow restic customCA to be from a configmap

tesshuflower:
Resolves: https://github.com/backube/volsync/issues/673

Updates spec to optionally allow for a configmap to be specified which contains the customCA rather than a secret.

Also moves some code around to hopefully be reusable if we add customCA capability to rclone for example.

Note that the role permissions needed to be updated to allow access to get/list/watch configmaps.

**Related issues:**
https://github.com/backube/volsync/issues/673


tesshuflower:
/cc @JohnStrunk 

FYI @birsanv

PR #708, opened 2023-04-04
Rclone customca

tesshuflower:

Updates Rclone functionality to allow specifying a customCA in the spec similar to what we do for restic.

**Related issues:**


birsanv (approved):
lgtm

Issue #653, opened 2023-03-01
Handling repo being stale locked for Restic based backup


budimanjojo:
**Describe the feature you'd like to have.**

Currently, when the `restic` replication pod got killed while performing job, the next pod will fail because the repo is still locked.
It would be nice if we can unlock the repo easily.

**What is the value to the end user? (why is it a priority?)**

It can fix jobs failing to complete because of locked repo resulting in an out of date state of the backups.

**How will we know we have a good solution? (acceptance criteria)**

Currently, the only way to remove locks is by doing `restic unlock` command manually using the CLI.
This is somewhat anti-pattern way of handling Kubernetes resources.
I am thinking about having an annotation/label for the `ReplicationSource` to instruct the operator to create a `Job` that does `restic unlock` command, something like `replicationsource.volsync.backube/remove-stale-lock: true`. Or even better we can give that annotation/label to the failing pod directly (is the operator watching for that resource?).

Another way to do this is by having `unlock` command built into the `volsync` kubectl plugin.

I'm fine with both implementations, or maybe there's better way to handle this, please do share your thought.

Thank you.

onedr0p:
Perhaps something could be built into the kubectl plugin to help cluster admins automate this process?

JohnStrunk:
We discussed this recently, and the current plan is to implement a solution via the ReplicationSource CR, allowing the user to indicate the desire to issue an unlock in a similar manner to how the "manual" trigger mechanism works.

### Proposal
There would be additional fields added to the `.spec` and `.status` of the RS:
```yaml
spec:
  restic:
    unlock: triggerString  #type: string
status:
  restic:
    unlock: responseString  #type: string
```

During a restic operation, if `.spec.restic.unlock` is non-empty and does not match `.status.restic.unlock`, prior to starting the backup action, a new "unlock" action would be run.
Once the Job completes, the `.status.restic.unlock` would be updated to match the string from the `.spec`.

This seems to meet the requirements for providing a way to unlock the repo (an imperative command). It doesn't require the user to install the CLI or restic locally, and they don't need to manipulate or read the restic repo Secret in the cluster.

tesshuflower:
/assign @tesshuflower 

@JohnStrunk I was considering using `.status.restic.lastUnlocked" instead of `unlocked` just to be consistent with some of our other status fields like `lastPruned` and `lastManualSync`.

tesshuflower:
Closing - implemented by PR #713

billimek:
[...]

PR #713, opened 2023-04-14
Restic unlock

tesshuflower:

Allows a `restic unlock` to be run before backup in the next restic sync (backup).  Only applicable to ReplicationSource.

The field in the spec is: `spec.restic.unlock`.  Once this is set to a unique string value, this will not trigger a sync, but on the next scheduled sync `restic unlock` will be run prior to the backup.  Afterrwards the status (`status.restic.lastUnlocked` will be updated to the same string value from `spec.restic.unlock`.  No further unlocks will be run unless `spec.restic.unlock` is set to a different value from the one in the status.

- Removing `spec.restic.unlock` entirely is supported, and will remove `spec.status.lastUnocked` from the status after the next sync (and will also not run a restic unlock).
- Originally I had intended to run unlock after the repo init, but it turns out this doesn't work in practice - `restic snapshots` itself will fail if an exclusive lock is held on the repo.  So instead I run the unlock first but catch the non-existent repo the same way we do in the init repo.  If no repo, then I assume no unlock is necessary.
- One interesting thing I found is that if the exclusive lock is created from another system, then the unlock can go through with no output or errors, but may not actually remove the lock.  I think this is due to lock created from another host, it doesn't get removed as it will not be considered stale yet.  So it's possible unlock will succeed (and not remove the lock) but the next backup will not.  Note that after 30 minutes, the lock will be considered stale and the unlock will remove it.  This doesn't appear to happen if the `restic unlock` happens from the same host that created the lock in the first place.
  - If this ever becomes an issue, it is possible to run `restic unlock --remove-all` which will remove all locks (the code is not doing this in the PR however)

**Related issues:**
https://github.com/backube/volsync/issues/653

### DIFF
--- a/api/v1alpha1/replicationsource_types.go
+++ b/api/v1alpha1/replicationsource_types.go
@@ -192,6 +192,15 @@ type ReplicationSourceResticSpec struct {
 	// CacheAccessModes can be used to set the accessModes of restic metadata cache volume
 	//+optional
 	CacheAccessModes []corev1.PersistentVolumeAccessMode `json:"cacheAccessModes,omitempty"`
+	// unlock is a string value that schedules an unlock on the restic repository during
+	// the next sync operation.
+	// Once a sync completes then status.restic.lastUnlocked is set to the same string value.
+	// To unlock a repository, set spec.restic.unlock to a known value and then wait for
+	// lastUnlocked to be updated by the operator to the same value,
+	// which means that the sync unlocked the repository by running a restic unlock command and
+	// then ran a backup.
+	// Unlock will not be run again unless spec.restic.unlock is set to a different value.
+	Unlock string `json:"unlock,omitempty"`
 	// MoverSecurityContext allows specifying the PodSecurityContext that will
 	// be used by the data mover
 	MoverSecurityContext *corev1.PodSecurityContext `json:"moverSecurityContext,omitempty"`
@@ -208,6 +217,10 @@ type ReplicationSourceResticStatus struct {
 	// lastPruned in the object holding the time of last pruned
 	//+optional
 	LastPruned *metav1.Time `json:"lastPruned,omitempty"`
+	// lastUnlocked is set to the last spec.restic.unlock when a sync is done that unlocks the
+	// restic repository.
+	//+optional
+	LastUnlocked string `json:"lastUnlocked,omitempty"`
 }
 
 // define the Syncthing field
--- a/controllers/mover/restic/restic_test.go
+++ b/controllers/mover/restic/restic_test.go
@@ -93,6 +93,60 @@ var _ = Describe("Restic retain policy", func() {
 	})
 })
 
+var _ = Describe("Restic unlock", func() {
+	var m *Mover
+	var owner *corev1.ConfigMap
+	logger := zap.New(zap.UseDevMode(true), zap.WriteTo(GinkgoWriter))
+	var start metav1.Time
+
+	BeforeEach(func() {
+		start = metav1.Now()
+		// The underlying type of owner doesn't matter
+		owner = &corev1.ConfigMap{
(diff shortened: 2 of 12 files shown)


JohnStrunk:
> One interesting thing I found is that if the exclusive lock is created from another system, then the unlock can go through with no output or errors, but may not actually remove the lock

An interesting point... What is considered "another system" by restic? Does a different pod or node cause it to think it's a different system or does [specifying the host name](https://github.com/backube/volsync/blob/39f34f539e0a5ab7df677954c5cdb9ff55de410d/mover-restic/entry.sh#L22) cause this not to be a problem for us?

tesshuflower:
> > One interesting thing I found is that if the exclusive lock is created from another system, then the unlock can go through with no output or errors, but may not actually remove the lock
> 
> An interesting point... What is considered "another system" by restic? Does a different pod or node cause it to think it's a different system or does [specifying the host name](https://github.com/backube/volsync/blob/39f34f539e0a5ab7df677954c5cdb9ff55de410d/mover-restic/entry.sh#L22) cause this not to be a problem for us?

I believe specifying the hostname like we do helps us here as I *think* it'll make it seem like the same host (which helps and means our unlock should unlock immediately without the 30 min wait on ones we've locked) - although I haven't been able to prove it - it's hard to artificially make restic in my env fail during prune so the lock gets left behind - I've been able to do it from a local restic where I kill the process while it's running prune - which is how I discovered this in the first place.  Even doing crtl-c on restic won't cause this as it cleans up properly if you interrupt it.  At first I thought it wasn't able to remove the lock altogether from a different "host".

api/v1alpha1/replicationsource_types.go
```diff
@@ -208,6 +217,10 @@ type ReplicationSourceResticStatus struct {
 	// lastPruned in the object holding the time of last pruned
 	//+optional
 	LastPruned *metav1.Time `json:"lastPruned,omitempty"`
+	// lastUnlocked is set to the last spec.restic.unlock when a sync is done that unlocks the
+	// restic repository.
+	//+optional
+	LastUnlocked string `json:"lastUnlocked,omitempty"`
```
budimanjojo, line 223: Shouldn't this also be of type `metav1.Time` instead of `string`?

tesshuflower: No, this field works more like the spec.trigger.manual string -User supplies a unique string `spec.restic.unlock` and then after a successful unlock the status.restic.lastUnlock is updated with the same string value.  The value can then be used to decide whether unlock needs to be run on the next sync (it will only run if the previous sync has a lastUnlocked that doesn't match).  See here: https://github.com/backube/volsync/issues/653#issuecomment-1508442673

controllers/mover/restic/restic_test.go
```diff
@@ -93,6 +93,60 @@ var _ = Describe("Restic retain policy", func() {
...
+			owner:        owner,
+			sourceStatus: &volsyncv1alpha1.ReplicationSourceResticStatus{},
+		}
+	})
+
+	When("Unlock is not set in the spec", func() {
+		It("shouldUnlock() should return false", func() {
+			Expect(m.shouldUnlock()).To(BeFalse())
+		})
+	})
```
JohnStrunk, line 123: Do you want to add a test case for `.spec` is "", but `.status` has a value?

I see it's covered in the job tests below.

JohnStrunk (commented):
/lgtm

tesshuflower:
We are currently working on a 0.8 release, but there are some features that need to go in first, and then possibly it'll be timed with a downstream release of ACM.  It's possible we may be able to release upstream ahead of this, but we'll need to make sure the main features (volume populator, and hopefully diskrsync) will be fully in and tested prior to cutting a release.

PR #727, opened 2023-04-25
Remove setting maxconcurrentreconciles dup

tesshuflower:

maxConcurrentReconciles seems to be set 2x in the replicationdestination controller SetupWithManager() function. I believe the intention was to set it to 100 but it gets set back to 10:

Log excerpt (see "worker count"):
```
2023-04-25T21:02:01.041Z	INFO	Starting workers	{"controller": "replicationdestination", "controllerGroup": "volsync.backube", "controllerKind": "ReplicationDestination", "worker count": 10}
2023-04-25T21:02:01.143Z	INFO	Starting workers	{"controller": "replicationsource", "controllerGroup": "volsync.backube", "controllerKind": "ReplicationSource", "worker count": 100}
```

**Related issues:**

### DIFF
--- a/controllers/replicationdestination_controller.go
+++ b/controllers/replicationdestination_controller.go
@@ -164,9 +164,6 @@ func (r *ReplicationDestinationReconciler) SetupWithManager(mgr ctrl.Manager) er
 		Owns(&rbacv1.Role{}).
 		Owns(&rbacv1.RoleBinding{}).
 		Owns(&snapv1.VolumeSnapshot{}).
-		WithOptions(controller.Options{
-			MaxConcurrentReconciles: 10,
-		}).
 		Complete(r)
 }
 


JohnStrunk:
The log messages have one for the source, and one for the destination. Are they different worker pools (i.e., per controller) or is there only one for the whole app?

tesshuflower:
> The log messages have one for the source, and one for the destination. Are they different worker pools (i.e., per controller) or is there only one for the whole app?

The settings seem to be per controller, both replicationsource and dest controllers have something like this:
https://github.com/backube/volsync/blob/afd6fdba462dbe5f4721d0ccfab303ffb0ef475d/controllers/replicationdestination_controller.go#L153-L158

It looks like there were some overlapping commits that both added `MaxConcurrentReconciles`, and then we ended up with it 2x in the destination (so it gets set to 100 and then to 10).  The source controller was fixed in this commit:  https://github.com/backube/volsync/commit/eeb950872492132ed5e5560e20d51c0a1f8fa4e9

JohnStrunk (commented):
/lgtm

PR #736, opened 2023-05-05
Added diskrsync to rsync ssh mover

awels:
Added diskrsync https://github.com/dop251/diskrsync for block devices to the ssh mover. This is to support syncing KubeVirt virtual machines, which primarily use block devices for the disk images.

diskrsync does not support TLS so I could not add it to the tls mover. If this is important, we will have to open a PR on diskrsync to add support for TLS.

This supports cloning block devices in an efficient manner, however right now only for the ssh mover.

**Related issues:**


shawn-hurley (commented):
Do we want to use the block device operations when a volume has a filesystem? 

we are mounting the device correctly, but I wonder if we want to only do this when the device is treated as a device in the original application.

awels:
We want to treat it as a block device, for us the main use case is VM disks, and thus the block device will most likely have some kind of filesystem on it, but we should not care about that. Also for mounting volumes inside the mover pods we would have to run as root at a minimum, and I am not sure we want that either.

awels:
Because we don't know what kind of filesystem (could be a windows VM for instance). And we don't want to do special handling when on our level it is presented to us as a block device. Also note that on the controller level we can't even identify if there is a filesystem on that block device.

shawn-hurley:
I don't have a huge dog in this fight, but I do want to understand the why.

If a volume is mounted as a VolumeMount, it always has a filesystem. If mounted as a device, it may or may not have a filesystem. 

 I am trying to understand why the default is to treat it like a block even when we can treat it like a filesystem because of the above rules.

tesshuflower:
> controllers/utils/utils.go

I think Shawn is bringing up a good point - it's possible the current rsync/rsync-tls would work for a block volume with a filesystem on it.  If it does work, we'd be breaking that by assuming we want to use diskrsync in those cases.

If that is the case (I might need to do some tests to validate), we might need to add some sort of opt-in to use block so we don't change the default current behaviour.

shawn-hurley:
and, if we want to break the default, what is the reason to do so IMO.

awels:
So there is no viable mechanism to identify if the block volume has a filesystem on it in the controller, so the controller has to make it a volumeDevice instead of a volumeMount. Then in the mover pod you could in theory check if there is a fileystem and use rsync. But you would require all kinds of privileges to do so. Or you could just treat it as a block device and use diskrsync.

tesshuflower:
I think we have these basic scenarios:

A) pvc with volumeMode Filesystem - works currently and is all we test against
B) pvc with volumeMode Block - if there is a filesystem on the disk, and we mount with volumeMount - then maybe this still works with current code
C) pvc with volumeMode Block - mounted in device mode we could use the new PR and use diskrsync

I think the concern is simply around B).  If this already works, then we'd be breaking this scenario by assuming any PVC with volumeMode block needs to be mounted in device mode and use diskrsync.  If B) does indeed work, then we may need to have a way to opt-in to using C).

awels:
For scenario B. The controller cannot know if there is a filesystem on the block device, and thus must create a mover job that creates a pod with a volumeDevice. Ergo the PVC will be visible as a block device in the container. Now assuming we can check if there is a compatible filesystem on the block, AND we are allowed to mount the filesystem (remember this is all normally done by kubernetes on a filesystem volume), then in theory we could potentially use just a regular rsync. But we would be creating a third option that will require a bunch of special handling, and special permissions. I personally think treating a block volume as a block volume regardless of what is on it, makes a lot more sense.

tesshuflower:
I think essentially we agree here - I don't want to try to auto-detect anything either at this point.  But we don't want to break something that could be working right now - I'd like to run a test at some point to see if B) even works as maybe this discussion is moot.

shawn-hurley:
If we were starting from the beginning, I could see that argument and maybe even could agree, but you are talking about making a change that is breaking what happens today, and that needs to be handled gracefully for the folks already using volsync.

awels:
What happens today is that block volume PVCs don't work at all. And you cannot mount block volumes as a volumeMount. If you try you get something like the following `volume example has volumeMode Block, but is specified in volumeMounts` and the pod doesn't start. So you HAVE to specify it in volumeDevices, which is not supported at all in any existing volsync release or code outside of this PR. So there is no possibility that scenario is out there.

shawn-hurley:
I am seeing this behavior, I would have sworn there is some case, where, you can use a block mode device, and k8s mounts it and adds a filesystem. maybe this is a CSI-specific implementation thing. 

I need to find this out, because I am getting this confused all over the place.

I am sorry, if this is the behavior across all versions of k8s then I am good with the change because it only handles this. 

I am sorry @tesshuflower and @awels, for misleading the conversation. If you don't mind giving me some period of time to try and find what I was talking about

awels:
So inside the CSI driver this is what happens when you ask for a filesystem volume, but the actual storage is backed by some block device. It creates a new block device, and puts a filesystem on it. This all happens inside the CSI driver and it has the appropriate privileges to do this. I do this in the CSI drivers I maintain. This is why I was confused about this since your container would have to have some impressive privileges to create a filesystem on the block device and mount it inside the container. And I do believe we want the movers to be completely unprivileged and rootless.

tesshuflower:
No problem! I learned something from all this at least :)

diskrsync-tcp/main.go
```diff
@@ -0,0 +1,273 @@
...
+		useReadBuffer = true
+	} else if !opts.noCompress {
+		sf, err := spgz.NewFromFileSize(f, os.O_RDWR|os.O_CREATE, diskrsync.DefTargetBlockSize)
+		if err != nil {
+			if err != spgz.ErrInvalidFormat {
+				if err == spgz.ErrPunchHoleNotSupported {
+					err = fmt.Errorf("target does not support compression. Try with -no-compress option (error was '%v')", err)
+				}
+				return err
+			}
```
tesshuflower, line 197: what happens here if the error is ErrInvalidFormat? Is this ignored on purpose?

awels: Yes it is, in that case it will assume it is just a raw file file and treat it as such.

tesshuflower:
/cc @JohnStrunk 

I think I'm good with this - do you want to review before we merge?
Notes: There are some block volume e2e tests using diskrsync, but have disabled them for non-openshift for the moment.

JohnStrunk:
Thanks @awels for sticking w/ this!

PR #789, opened 2023-07-05
Fips fixes

tesshuflower:

- Restrict metrics to TLS 1.2 (was TLS 1.3).  TLS 1.3 has issues with FIPS envs in older OpenShift envs
- also USE CGO_ENABLED=1 when building.  This is really only required when using the openshift go toolset to build (for FIPS support) - but changing here so we're consistent with downstream builds.

See: https://access.redhat.com/security/vulnerabilities/RHSB-2023-001

**Related issues:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ WORKDIR /workspace
 # We don't vendor modules. Enforce that behavior
 ENV GOFLAGS=-mod=readonly
 ENV GO111MODULE=on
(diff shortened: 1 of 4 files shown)


tesshuflower:
Will need to pull into the release-0.7 stream

/cherry-pick release-0.7

JohnStrunk (commented):
/lgtm

PR #796, opened 2023-07-12
restic: Make explicit sync only happen after restore

### LINKED ISSUE #788: Restic backup never reaches Done even though Restic completed successfully
**Describe the bug**
Restic backup never reaches `=== Done ===` even though Restic completed successfully

**Steps to reproduce**
Take a restic backup

**Expected behavior**
finish the backup

**Actual results**
```
Starting container
VolSync restic container version: v0.7.1+8af0bf0
backup
restic 0.15.1 compiled with go1.19.7 on linux/amd64
Testing mandatory env variables
== Checking directory for content ===
== Initialize Dir =======
ID        Time                 Host        Tags        Paths
------------------------------------------------------------
88c570d9  2023-07-04 06:51:26  volsync                 /data
------------------------------------------------------------
1 snapshots
=== Starting backup ===
/data /
using parent snapshot 88c570d9

Files:           0 new,   271 changed,     0 unmodified
Dirs:            0 new,    15 changed,     0 unmodified
Added to the repository: 98.609 KiB (21.898 KiB stored)

processed 271 files, 19.028 MiB in 0:00
snapshot 58391a18 saved
/
=== Starting forget ===
Applying Policy: keep 10 daily snapshots and all snapshots within 10d of the newest
keep 2 snapshots:
ID        Time                 Host        Tags        Reasons         Paths
----------------------------------------------------------------------------
88c570d9  2023-07-04 06:51:26  volsync                 within 10d      /data
58391a18  2023-07-04 06:58:26  volsync                 within 10d      /data
                                                       daily snapshot
----------------------------------------------------------------------------
2 snapshots

Restic completed in 2s


```

**Additional context**
I have a few dozen restic backups, some of which fail like this with the pod stuck running because the restic backup never reaches the `Done` status. There is nothing special for the `RS`-es where it works vs where it doesn't. As a matter of fact I recreated them and pointed them to new restic repos and they seem random (some work some don't).

Example of a `ReplicationSource` where this happens:

```
---
apiVersion: volsync.backube/v1alpha1
kind: ReplicationSource
metadata:
  name: calibre-web
  namespace: media
spec:
  sourcePVC: config-calibre-web-0
  trigger:
    schedule: "0 */12 * * *"
  restic:
    copyMethod: Snapshot
    pruneIntervalDays: 14
    repository: calibre-web-restic-secret
    moverSecurityContext:
      runAsUser: 568
      runAsGroup: 568
      fsGroup: 568
    retain:
      daily: 14
```

**I would like to add that this is random new behavior, the exact same RS-es worked fine at one point and there were no changes made on any side.**

JohnStrunk:
This moves the `sync` call at the end of the restic mover script so that it only happens after restore operations.

The sync exists to ensure data is flushed prior to creating a Snapshot. It is needed because we have no guarantee of sequencing between the node's `unmount` and the CSI snapshot operation that is triggered by our creation of the Snapshot object. Since the snapshot would only affect the destination volume (restore), it is only applicable there.
The unmount of volumes will cause their data to eventually get flushed, so when we don't care about sequencing, we don't need the explicit sync.

**Related issues:**
Fixes #788 
Related #792

### DIFF
--- a/mover-restic/entry.sh
+++ b/mover-restic/entry.sh
@@ -162,14 +162,14 @@ function reverse_array() {
 }
 
 ################################################################
-# Selects the first restic snapshot available for the 
-# given constraints. If RESTORE_AS_OF is defined, then 
+# Selects the first restic snapshot available for the
+# given constraints. If RESTORE_AS_OF is defined, then
 # only snapshots that were created prior to it are considered.
-# If SELECT_PREVIOUS is defined, then the n-th snapshot 
-# is selected under the matching criteria. 
-# If a snapshot satisfying the conditions is found, then its ID 
+# If SELECT_PREVIOUS is defined, then the n-th snapshot
+# is selected under the matching criteria.
+# If a snapshot satisfying the conditions is found, then its ID
 # is returned.
-# 
+#
 # Globals:
 #   SELECT_PREVIOUS
 #   RESTORE_AS_OF
@@ -186,7 +186,7 @@ function select_restic_snapshot_to_restore() {
     local snapshot_id
     local snapshot_ts
     local trimmed_timestamp
-    local snapshot_epoch 
+    local snapshot_epoch
 
     # go through the timestamps received from restic
     IFS=$'\n'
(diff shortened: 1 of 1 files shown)


tesshuflower (approved):
/lgtm

Issue #557, opened 2022-12-13
Allow setting keep last on restic data mover


onedr0p:
**Describe the feature you'd like to have.**

According to the [Restic docs](https://restic.readthedocs.io/en/stable/060_forget.html#removing-snapshots-according-to-a-policy) there is a `--keep-last` which will keep the latest number of snapshots.

It would be nice if this was able to be used in a `ReplicationDestination`. e.g.

```yaml
---
apiVersion: volsync.backube/v1alpha1
kind: ReplicationSource
metadata:
   name: database-source
   namespace: source
spec:
   sourcePVC: mysql-pv-claim
   trigger:
      schedule: "*/30 * * * *"
   restic:
     pruneIntervalDays: 15
     repository: restic-config
     retain:
       # Last will only keep the last X number of snapshots
       last: 14
     copyMethod: Clone
```

**What is the value to the end user? (why is it a priority?)**

Allows backups to be retained on a number of snapshots level, not time based

**How will we know we have a good solution? (acceptance criteria)**

**Additional context**

lltr:
Seems like the docs at https://volsync.readthedocs.io/_/downloads/en/latest/pdf/, page 107 shows an example of using `last` in a 
`ReplicationSource`, can I confirm if this issue is completed?

onedr0p:
For now I am able to use `within` as a workaround.

```yaml
    retain:
      daily: 7
      within: 3d
```

This will keep all snapshots within 3 days and also 1 per day for up to 7 days.

onedr0p:
The PR was merged, thanks all

Some docs need to be updated. https://github.com/backube/volsync/issues/801

PR #797, opened 2023-07-12
feat: expose `last` on `retain` options for restic mover

onedr0p:

Allows users to set `last` as forget option in the ReplicationDestination  CR.

I believe I hit all the right spots in the code as the functionality was mainly there just not exposed as an option.

**Related issues:**

https://github.com/backube/volsync/issues/557


tesshuflower:
@JohnStrunk what do you think about allowing restic keep-last?  Let me know if you're ok with this one and I could help add a unit test as well.

JohnStrunk:
My concern with this is if the backup portion succeeds, but something else (like the forget) fails such that the operator thinks the container failed. This can cause the operator to loop, retrying backups. The result could end up being a large number of backups in rapid succession.
"keep last" would cause the older backups to get purged, leaving the user with all backups from approximately the same time.
This is why I didn't enable it originally.

Having said that, the current structure of the restic script makes that fairly unlikely since all it does is a forget right after the backup.

Honestly, I'm on the fence w/ whether to take this because it's dangerous for the user, in a way that isn't obvious.

JohnStrunk:
This also needs a `make generate` added to the commit to fix the failed test

tesshuflower:
fixed the make generate, but I do think @JohnStrunk your concerns are valid about potentially losing older snapshots unintentionally.

onedr0p:
> This can cause the operator to loop, retrying backups. The result could end up being a large number of backups in rapid succession.

I've seen then happen on my end but maybe it's worth putting in the docs to use `last` only with another type like `daily` so that at least if the backup loop happens at least `daily` will be retained?

onedr0p:
@JohnStrunk with that said if `last` is supported and the backup loop happens at last it won't flood the restic repo with hundreds or more snapshots.

PR #803, opened 2023-07-15
chore: add `.editorconfig`

onedr0p:

Most editors support this file and it's a good way that enforces some styling for contributors.


onedr0p:
Not stale

onedr0p:
Not stale, this would be a great QoL change.

Issue #627, opened 2023-02-19
Option to NOT run backup as soon as the replicationsource is applied to the cluster


onedr0p:
**Describe the feature you'd like to have.**

Hi 👋🏼 

Please add a configuration option to not have the backups run when the replicationsource is applied to the cluster.

**What is the value to the end user? (why is it a priority?)**

When bootstrapping a new cluster using GitOps the replicationsource is applied and will start backing up data right away. This isn't really ideal since I want to recover from a previous backup.

**How will we know we have a good solution? (acceptance criteria)**

An option is added to prevent backups being taken as soon as a replicationsource is applied to the cluster.

**Additional context**

budimanjojo:
Maybe implementing Point In Time Recovery will solve this problem too if this can be implemented in Volsync. But it requires the method to supports incremental backups (I believe `restic` does?) So we can restore the data using `timestamp` instead of specifying which backup to restore.

But I'm fine with having option not to backup on creation like this issue describes too if this is too hard to implement.

tesshuflower:
@onedr0p If you're looking for something you can do right now - you could take a look at spec.paused = true in the replicationsource spec.

Something like:

```
apiVersion: volsync.backube/v1alpha1
kind: ReplicationSource
metadata:
  name: my-repl-source
spec:
  paused: true
  sourcePVC: mypvc
...
```

What this will do is still create a job for the replicationsource backup, however the job will run with parallelism = 0 so no actual pod will run until you remove `paused` from the spec or set it to false.

However, be aware that if you're using a copyMethod of Snapshot or Clone, it means the snapshot/clone will still be taken immediately - so when you do unpause later on, this first backup will still be of the (empty?) pvc when you initially created the replicationsource.

onedr0p:
@tesshuflower that isn't really all that ideal because I would have to edit a ton of ReplicationSources in a GitOps repo and have `paused: true` when restoring my clusters and then updating them all again to `paused: false` when the clusters are bootstrapped and apps are ready to be restored. And as you said since I am using `Snapshot` it would still create a back of stale data which leads into issue https://github.com/backube/volsync/issues/555

I am curious if there could be a global option for the controller to handle this?

```yaml
  values:
    # When a ReplicationSource is created run a backup immediately
    backupImmediately: false
    manageCRDs: true
    metrics:
      disableAuth: true
```

JohnStrunk:
[...]

PR #812, opened 2023-07-27
feat: Add immediate option to `ReplicationSource`

onedr0p:

Adds immediate option to `ReplicationSource` - immediate can be used to stop replication when a new ReplicationSource is created. Defaults to "true".

I am unable to run `make test` (Apple Silicon issue?) locally so looking for some feedback. This also could use some unit tests if required, however I am unable to write & test due to that issue until I find another way.

**Related issues:**
https://github.com/backube/volsync/issues/627


controllers/mover/restic/builder.go
```diff
@@ -119,6 +119,8 @@ func (rb *Builder) FromSource(client client.Client, logger logr.Logger,
 	saHandler := utils.NewSAHandler(client, source, isSource, privileged,
 		source.Spec.Restic.MoverServiceAccount)
 
+	isImmediate := source.Spec.Immediate == nil
```
uhthomas, line 122:
```suggestion
	isImmediate := source.Spec.Immediate != nil
```

?

uhthomas: Isn't this variable a boolean, why is it being compared to nil?

onedr0p: The instantiation of `Immediate` is a pointer, so here I am making the default `true`. I couldn't see another way to do this so I am all ears if there is a better way.

uhthomas: I see, but the actual value is never consumed anywhere. It should be something like:

```suggestion
	isImmediate := true
	if source.Spec.Immediate != nil {

		isImmediate = * source.Spec.Immediate

	
}
```

tesshuflower:
@onedr0p thanks for this PR, however I think if we do this there are a few updates we'd need to do:

- If we add immediate to replicationsource I'd prefer to also add to replicationdestination to be consistent
- Does it make sense to add `immediate` to spec.trigger somewhere? And potentially even call this something like `spec.trigger.scheduleImmediate`?  Since I think this will not apply (and should be ignored) by manual triggers.
- I think the actual implementation will need to be modified to set status/modify the scheduling algorithm to make sure we skip this initial sync if immediate is false.  The way it's done in the PR prevents the job from starting - which would be the same effect as creating a replicationsource with paused=true.  The problem with this is everything else will still happen immediately - i.e. snapshot/clone will be taken of the pvc immediately, and that pvc from snap/clone will be of the PVC at the time when the replicationsource is created.  I think to do this as you intend, we'd need to modify the scheduling to not sync at all at the start which is a little more involved as we don't want to risk breaking the scheduling.  

I'll discuss with the rest of the team to see if this does sound like a valid direction to go - I will try to take a look at the scheduling when I have a moment and see if it's something that could be done relatively easily.

onedr0p:
Closing PR, will cycle back if needed after the volume populator is released.

PR #824, opened 2023-08-14
Use YQ v4.34.2 because latest (v4.35.1) requires golang 1.20

tesshuflower:
- lock yq to v4.34.2 because the latest yq requires golang 1.20
- This was preventing unit tests from passing in PRs

this can be rolled back after we update to golang 1.20

**Related issues:**


JohnStrunk (commented):
/lgtm

PR #848, opened 2023-08-31
Fix cronspec validation

### LINKED ISSUE #847: The kubebuilder validation for replication source is too restrictive
**Describe the bug**
I attempted to create a volsync cron schedule of 0 9-17 * * *, and this was not allowed as it didn't fit with the validation rule inserted here.
https://github.com/backube/volsync/blob/main/api/v1alpha1/replicationsource_types.go#L49

**Steps to reproduce**
Attempt to use a more complicated cron expression than numbers and *s

**Expected behavior**
You can define complex cron expressions

**Actual results**
Kubernetes did not accept the value, because it failed validation

**Additional context**

JohnStrunk:
Improves the regex validation in the CRD for the schedule cronspec

Shellcheck alerted on a couple other issues which are fixed in a separate commit

**Related issues:**
Fixes #847


tesshuflower (approved):
/lgtm

Thanks John!

PR #849, opened 2023-08-31
Generate server certs at startup if first time

tesshuflower:
- Will leave cert.pem, key.pem in the syncthing config dir if they're already there

- This fixes issues with cert generation at startup in fips envs

**Related issues:**


JohnStrunk (commented):
/lgtm

PR #871, opened 2023-09-18
Switch to pre-commit for linting

JohnStrunk:
This adds [pre-commit](https://pre-commit.com/) as the main linter in place of the old shell script (and associated dependencies).

I have swapped it out in CI, and it can be installed locally as a git pre-commit hook (its real intended function)
```console
$ pip install --user --upgrade pre-commit
$ pre-commit install -f
```

All the old linters should still be there, but run via pre-commit now, plus some more.

**Related issues:**


JohnStrunk:
/retest
Failed during deprovisioning... that shouldn't be a thing.

PR #875, opened 2023-09-20
git clone and run go build for diskrsync

tesshuflower:
- Will help with downstream builds so downstream can be as close to upstream as possible
- Also update to set a version (and print it out in the log) for diskrsync-tcp

**Related issues:**

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -49,6 +49,15 @@
       ]
     },
     {
+      // Group diskrsync and spgz dependent library Docker and go updates
+      // together in a single PR
+      "groupName": "diskrsync version",
+      "groupSlug": "diskrsync",
+      "matchPackagePrefixes": [
+        "github.com/dop251/"
+      ],
+    },
+    {
       // Renovate's packages update too frequently, so we only schedule updates
       // once a week to keep the noise down
       "matchPackageNames": ["renovatebot/pre-commit-hooks"],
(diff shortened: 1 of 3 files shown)


JohnStrunk:
Looks good. Do you want to try updating the renovate config to do auto-updates? It should be just like rclone/syncthing

tesshuflower:
@JohnStrunk I attempted to modify the renovate config - I think the dockerfile update for diskrsync should be ok, but I also tried to add a grouping  but not entirely sure it'll actually group the dockerfile updates with the go updates with this (not sure how to test either).  Do you mind looking at this in particular?

https://github.com/backube/volsync/blob/4e8be0847a513ccb5e42d6d9d2dee1f270f6bf5e/renovate.json5#L52-L59

I think ultimately I'd prefer to leave that spgz library at whatever version diskrsync at our specific version requires, but that is even more complex - For now I'm just trying to group the updates together in a single PR (spgz and diskrsync go module and diskrsync in the Dockerfile)

renovate.json5
```diff
@@ -48,6 +48,15 @@
...
     },
+    {
+      // Group diskrsync and spgz dependent library Docker and go updates
+      // together in a single PR
+      "groupName": "diskrsync version",
+      "groupSlug": "diskrsync",
+      "matchPackagePrefixes": [
+        "github.com/dop251/"
+      ],
+    },
```
JohnStrunk, line 59:
```suggestion
    {
      // Group diskrsync and spgz dependent library Docker and go updates
      // together in a single PR
      "groupName": "diskrsync packages",
      "groupSlug": "diskrsync",
      "matchPackagePatterns": [
        // Using the "packagePatterns" option so that we can match both the go
        // packages (w/o the "github.com/" prefix) and the other references that
        // do have it.
        "dop251/diskrsync",
        "dop251/spgz"
      ]
    },
```

The above is what I had to do to group all the instances of diskrsync and spgz together

PR #891, opened 2023-09-25
Remove importlib-resources

JohnStrunk:
See if we can remove the importlib dependency which pipenv cannot solve for.
importlib was added to allow Ansible to run w/ an older version of Python
See: b5e611f10

**Related issues:**


tesshuflower:
@JohnStrunk I'm going to make a PR to update the custom-scorecard-tests image so it uses a newer python version (just trying to build it and then confirm it runs successfully first).

Currently the custom-scorecard-tests image is using a ubi9-minimal img  which packages python 3.9 and I think it's having issues with importlib-resources and the Pipfile.lock in test-e2e - it may cause this PR to fail until I fix it.

PR #897, opened 2023-09-26
e2e tests - remove importlib-resources dependency

tesshuflower:
- Dependency is causing update issues and only needed for python 3.9 and earlier. Github actions are using python 3.10 for e2e tests and custom scorecard tests image is now using python 3.11

**Related issues:**


tesshuflower:
Prow e2e tests are still using a ubi9 base image (which will package python 3.9) - Created PR in openshift/release to see if this can be updated to use python 3.11 like what we do in the custom-scorecard-tests image:  https://github.com/openshift/release/pull/43732

PR #898, opened 2023-09-26
test-e2e-install target, make --user flag optional

tesshuflower:

Trying to update our CI prow jobs to use a base image with python-311 here: https://github.com/openshift/release/pull/43732.  (this uses image registry.access.redhat.com/ubi9/python-311).

However this seems to start a virtual env automatically?  So when running "make test-e2e-install", it runs:

```
./.ci-scripts/retry.sh pip install --user --upgrade pipenv==2023.9.8
```

And this fails with an error like this:
```
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
```

This change simply makes the --user flag a setting via `PIP_INSTALL_ARGS`.  This way in the prow job we can override to install without the `--user` flag by running:

```
make test-e2e-install PIP_INSTALL_ARGS=
```

**Related issues:**


JohnStrunk (commented):
/lgtm

PR #914, opened 2023-10-03
Renovate: Group syncthing updates into single PR

JohnStrunk:

**Related issues:**

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -62,6 +62,17 @@
       ]
     },
     {
+      // Group syncthing updates together in a single PR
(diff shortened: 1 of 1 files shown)


tesshuflower:
> /cc @tesshuflower
> 
> ... assuming you think upgrading both at the same time is a good idea. I know we've had cases where we only backport the client lib.

Yes, this looks good to me - the backport updates were usually due to things like CVEs where we needed to update go modules but we didn't want to update .Y versions of syncthing itself in older releases.  For main I agree this is nice to have both syncthing server + client libs updated together 👍

PR #984, opened 2023-11-14
ci: Limit the amount of parallelism in e2e tests

JohnStrunk:
It seems like we've been having a lot of intermittent CI failures. This switches from running a fixed number of "batches" to capping the maximum parallelism. The batch size was increasing as additional tests were added... By switching to parallelism limits, the number of batches will automatically increase as needed.

**Related issues:**

### DIFF
--- a/.github/workflows/operator.yml
+++ b/.github/workflows/operator.yml
@@ -377,7 +377,7 @@ jobs:
 
       - name: Run e2e tests
         env:
-          BATCHES: 2
+          MAX_PARALLELISM: 5
         run: make test-e2e
 
   # This is a dummy job that can be used to determine success of CI:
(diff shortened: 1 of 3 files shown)


JohnStrunk:
Even with the decreased parallelism, the Syncthing tests are still failing intermittently. The issue seems to be that it's taking too long to establish a connection to the local instance's API:
```
        "Events:",
        "  Type    Reason                        Age    From                Message",
        "  ----    ------                        ----   ----                -------",
        "  Normal  PersistentVolumeClaimCreated  13m    volsync-controller  created PersistentVolumeClaim/volsync-test-2-config to receive incoming data",
        "  Normal  DaemonConnected               4m40s  volsync-controller  Connection established to local Syncthing API"
    ]
```

tesshuflower:
Wonder how tiny these envs are... the syncthing e2e tests do run the most pods, since they have 3 namespaces with 1 syncthing each.  Possible some of those pods are not able to start in time?  Especially if the 2 syncthing e2e tests are running in parallel?

JohnStrunk:
Reviving this to see if it helps
/cc @tesshuflower

PR #1000, opened 2023-11-26
support for restricted PSS

ajgon:
When trying to do a `ReplicationDestination` for volumes in PSS restricted namespace, job is not run because of missing `containers.runAsRoot` stanza:

```
Error creating: pods "volsync-dst-mypod-myvol-lzkx5" is forbidden: violates PodSecurity "restricted:latest": runAsNonRoot != true (pod or container "restic" must set securityContext.runAsNonRoot=true)
```

This PR attempts to fix it, without breaking backwards compatibility. It does it simply by propagating `moverSecurityContext.runAsNonRoot` from pod context to containers context.

I'm using restic as my backup tool of choice, I built a image with propsed change and can confirm it works. Also added tests where applicable, and they also seem to be passing.

**Related issues:**
None ?

### DIFF
--- a/controllers/mover/rclone/mover.go
+++ b/controllers/mover/rclone/mover.go
@@ -298,6 +298,10 @@ func (m *Mover) ensureJob(ctx context.Context, dataPVC *corev1.PersistentVolumeC
 
 		podSpec := &job.Spec.Template.Spec
 
+		if m.moverSecurityContext != nil && m.moverSecurityContext.RunAsNonRoot != nil {
+			podSpec.Containers[0].SecurityContext.RunAsNonRoot = m.moverSecurityContext.RunAsNonRoot
+		}
+
 		if customCAObj != nil {
 			// Tell mover where to find the cert
 			podSpec.Containers[0].Env = append(podSpec.Containers[0].Env, corev1.EnvVar{
--- a/controllers/mover/restic/mover.go
+++ b/controllers/mover/restic/mover.go
@@ -409,6 +409,9 @@ func (m *Mover) ensureJob(ctx context.Context, cachePVC *corev1.PersistentVolume
 				{Name: "tempdir", MountPath: "/tmp"},
 			},
 		}}
+		if m.moverSecurityContext != nil && m.moverSecurityContext.RunAsNonRoot != nil {
+			podSpec.Containers[0].SecurityContext.RunAsNonRoot = m.moverSecurityContext.RunAsNonRoot
+		}
 		podSpec.RestartPolicy = corev1.RestartPolicyNever
 		podSpec.ServiceAccountName = sa.Name
 		podSpec.SecurityContext = m.moverSecurityContext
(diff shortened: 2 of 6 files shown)


tesshuflower:
Hey thanks for this @ajgon !  Do you have any more info about the distro you're using where you see this issue?  Normally I thought runAsNonRoot wasn't needed at the container level as it should be determined from the pod level if not set - but maybe that assumption is wrong.

ajgon:
Hey @tesshuflower - I'm using [Talos Linux](https://www.talos.dev/) with most recent (1.28.4) version of kubernetes. Maybe something changed recently?

ajgon:
Hard to tell because I started testing this when I already had k8s 1.28. But I can spin test cluster later today with 1.27 and see if it works.

tesshuflower:
@ajgon If you can recreate, could you attach the yaml for your replicationdestination, as well as getting the yaml for the job & pod started for the volsync mover?

ajgon:
@tesshuflower Hey, sorry for late reply, got weekend busier than I expected :-). 

Looks like, this is something new in k8s 1.28.x. I tested restore with the same manifests on my testing 1.27.8 cluster, on restricted namespace and everything went smoothly.

Interestingly I can't find anything about it in k8s changelog, so not sure if it was sneaked out with another commit or fixed alongside with another bug ¯\_ツ_/¯

tesshuflower:
Wonder if something unusual is going on then - it might be worth capturing the job & pod yaml when you encounter this error.

According to the docs: https://kubernetes.io/docs/concepts/security/pod-security-standards/ the container field for runAsNonRoot shouldn't need to be set.  Look under "Running as non-root"

```
The container fields may be undefined/nil if the pod-level spec.securityContext.runAsNonRoot is set to true.
```

So perhaps something else is mutating the job or pod and setting the container fields unexpectedly?

I don't see anything obvious in the pod security admission function that would cause it to fail if the container field is not set, but the pod field is:  https://github.com/kubernetes/pod-security-admission/blob/1c6715f1f1adccb92735cb4dfc3bfdddb3177473/policy/check_runAsNonRoot.go#L61

ajgon:
Ok, that makes a lot of sense. I'll do some more extensive testing tomorrow and let you know about my findings - I have few moving parts in my cluster, maybe something interferes 🤔 I'll spin up dummy 1.28.4 cluster and see if it helps.

ajgon:
Ok I'm completely lost, now it works :D Maybe during my initial tests I didn't set up `runAsNonRoot: true` on `moreSecurityContext`, and did it just after applying my patch to volsync. So had wrong impression that patching helped, while I never set the variable in a first place 🤔 

Anyway, you are completely right - setting `runAsNonRoot` only on pod level is sufficient enough. So my PR is now only matter of question - does it make sense to propagate it to container level? I leave the decission to you (abeit now I think of it, it doesn't add much value) - if it doesn't, this PR can be closed.

Thanks for this amazing tool, and keep doing a good job 💪

tesshuflower:
Thanks for the additional info @ajgon !

I'm going to close this PR, I think it's not necessary to change right now - we can always re-evaluate if this issue does come up again.

PR #1046, opened 2023-12-20
PRs that have "hold" label should not be marked stale

JohnStrunk:

**Related issues:**


tesshuflower:
@JohnStrunk This PR seems to be stuck, not sure how to get it to re-run the failed jobs - what do you think about force-merging it?  We don't really need the e2e to run anyway for this change.

JohnStrunk:
done :smile_cat:

Issue #707, opened 2023-04-04
Customize cpu/Memory resourceRequirements/Limit on mover pods


kaovilai:
**Describe the feature you'd like to have.**

Customize from default mover pod cpu/memory resourceRequirements/resourceLimits
**What is the value to the end user? (why is it a priority?)**

Ability to scale pod requirements to the workload of the pod, ie very large, or very small.
**How will we know we have a good solution? (acceptance criteria)**

cpu/memory resourceRequirements/resourceLimits can be set
**Additional context**

josephassiga:
Hello team, is there any way to advance this issue, it is hihgly rated by ou customer

tesshuflower:
I'll try to see if we can prioritize this in the v0.9 timeframe

JohnStrunk:
Since this seems to be getting some attention, I'll add some thoughts...

As we look at setting [requests and limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/), there are significant implications of doing so. In the case of **requests**, I think it's a net benefit since it provides a way for the user to effectively reserve both CPU and memory for the mover pods so that they can be scheduled to a node where they will run efficiently and not risk OOM-kill. The down-side is that this makes it harder to schedule the mover pods, given that they may already be restricted due to storage topology. A resource request may make them unschedulable.

In the case of **limits**, while it makes sense for us to add them by bringing in the appropriate go struct from kube, I don't think I would recommend that end users set them. By limiting CPU, it will cause replication to take longer with basically no up-side assuming resource requests for important workloads are properly set. Setting limits on memory will cause the replication to fail if the limit is hit. Unfortunately, we have no way to quantify the expected memory consumption of a given mover (*looking at you, Restic*), so even using this as a safety valve can be problematic.

igor-nikiforov:
We can't use volsync without this due LimitRanges forced on namespace.

@tesshuflower I see that https://github.com/backube/volsync/pull/1072 was merged but changes still not available in latest helm chart release. Could you please advise any ETA when it will be available in helm chart? Thanks.

tesshuflower:
@igor-nikiforov good timing as we are actually trying to get our v0.9.0 release (which will contain this feature) done - I'm hoping this will happen today sometime, possibly tomorrow.

PR #1054, opened 2024-01-04
feat: Add podLabels to Helm chart

bjw-s:
It is already possible to define Pod annotations on the volsync controller Deployment. This PR makes it possible to set Pod labels as well.

**Related issues:**

### DIFF
--- a/helm/volsync/Chart.yaml
+++ b/helm/volsync/Chart.yaml
@@ -20,7 +20,7 @@ keywords:
   - disaster recovery
   - replication
   - storage
-annotations:  # https://artifacthub.io/docs/topics/annotations/helm/
+annotations: # https://artifacthub.io/docs/topics/annotations/helm/
   # Changelog for current chart & app version
   # Kinds: added, changed, deprecated, removed, fixed, and security
   artifacthub.io/changes: |
@@ -66,7 +66,7 @@ kubeVersion: "^1.20.0-0"
 # This is the chart version. This version number should be incremented each time
 # you make changes to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: "0.8.0"
+version: "0.9.0"
 
 # This is the version number of the application being deployed. This version
 # number should be incremented each time you make changes to the application.
--- a/helm/volsync/Chart.yaml
+++ b/helm/volsync/Chart.yaml
@@ -66,7 +66,7 @@ kubeVersion: "^1.20.0-0"
 # This is the chart version. This version number should be incremented each time
 # you make changes to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: "0.9.0"
+version: "0.8.0"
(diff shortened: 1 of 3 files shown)


helm/volsync/Chart.yaml
```diff
@@ -66,7 +66,7 @@ kubeVersion: "^1.20.0-0"
 # This is the chart version. This version number should be incremented each time
 # you make changes to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: "0.8.0"
```
tesshuflower, line 69: Can you roll back this change? I don't want to modify the default version to v0.9.0 at this point

bjw-s: I'm fine doing that, I merely changed it because of that instruction `This version number should be incremented each time you make changes to the chart` following semVer.

I will submit a PR soon.

bjw-s: I've removed the chart version bump.
Looks like some formatter fixes were applied automatically, are they OK to keep?

tesshuflower: Thanks for that - you're right normally we would update the version, but we'll do it at the same time when we make a v0.9 release.  I think the formatter fixes are probably ok.

tesshuflower:
@JohnStrunk are you ok with this update to the helm charts?

JohnStrunk:
Looks fine to me. Of note: this doesn't cover the mover pods, just the VolSync controller.
/ok-to-test
/approve

bjw-s:
> Looks fine to me. Of note: this doesn't cover the mover pods, just the VolSync controller.
> 
> /ok-to-test
> 
> /approve

I would love to be able to set labels on the mover pods as well (to more easily target them with networkPolicies for example), but that would require changes to the Go code. I haven't dared to dig in to that yet 😅

tesshuflower:
> > Looks fine to me. Of note: this doesn't cover the mover pods, just the VolSync controller.
> > /ok-to-test
> > /approve
> 
> I would love to be able to set labels on the mover pods as well (to more easily target them with networkPolicies for example), but that would require changes to the Go code. I haven't dared to dig in to that yet 😅

@bjw-s FYI: We've just made a change for https://github.com/backube/volsync/issues/707 which also included the ability to set podLabels for a mover.  See PR: https://github.com/backube/volsync/pull/1072

bjw-s:
> @bjw-s FYI: We've just made a change for https://github.com/backube/volsync/issues/707 which also included the ability to set podLabels for a mover.  See PR: https://github.com/backube/volsync/pull/1072

That's awesome! Thank you! 🙏

bjw-s:
@tesshuflower I'm having some trouble figuring out where the test is failing exactly (and how it can be failing based on my changes 😅). Can you perhaps offer some guidance?

tesshuflower:
> Details

Most likely not due to your change, sometimes we're still getting intermittent failures as these test clusters are pretty resource constrained.

/retest

PR #1080, opened 2024-02-01
Copy trigger annotations

tesshuflower:

Allows copy-trigger annotations on the pvc to pause/trigger snapshots or clones in a sync

- latestMoverStatus is handled a little strangely - if we timeout waiting on someone to update the `copy-trigger` I will put the error and status in `latestMoverStatus`.  This was normally reserved for only errors during a job, but seemed like the best place to put this.  If we just return an error it'll end up in the conditions - but then will get overwritten as soon as we try to sync again with a syncing  status.  I think this is the case in general with errors (that are not job failures) in VolSync.

**Related issues:**

### DIFF
--- a/controllers/utils/copy_trigger_annotations.go
+++ b/controllers/utils/copy_trigger_annotations.go
@@ -1,0 +1,145 @@
+/*
+Copyright 2024 The VolSync authors.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published
+by the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+package utils
+
+import (
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// Annotation optionally set on src pvc by user.  When set, a volsync source replication
+	// that is using CopyMode: Snapshot or Clone will wait for the user to set a unique copy-trigger
+	// before proceeding to take the src snapshot/clone.
+	UseCopyTriggerAnnotation = "volsync.backube/use-copy-trigger"
+	SnapTriggerAnnotation    = "volsync.backube/copy-trigger"
+
+	// Annotations for status set by VolSync on a src pvc if UseCopyTriggerAnnotation is set to "true"
+	LatestCopyTriggerAnnotation             = "volsync.backube/latest-copy-trigger"
+	LatestCopyStatusAnnotation              = "volsync.backube/latest-copy-status"
+	LatestCopyTriggerWaitingSinceAnnotation = "volsync.backube/latest-copy-trigger-waiting-since"
+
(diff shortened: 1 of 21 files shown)


controllers/utils/copy_trigger_annotations.go
```diff
@@ -0,0 +1,134 @@
...
+	LatestCopyStatusAnnotation              = "volsync.backube/latest-copy-status"
+	LatestCopyTriggerWaitingSinceAnnotation = "volsync.backube/latest-copy-trigger-waiting-since"
+
+	// VolSync latest-copy-status annotation values
+	LatestCopyStatusValueWaitingForTrigger = "WaitingForTrigger"
+	LatestCopyStatusValueInProgress        = "InProgress"
+	LatestCopyStatusValueCompleted         = "Completed"
+
+	CopyTriggerWaitTimeout time.Duration = 10 * time.Minute
+)
```
JohnStrunk, line 45: Let's move these out to `api/v1alpha1/common_types.go` since users and other controllers are expected to use these "well-known" names.

controllers/utils/copy_trigger_annotations.go
```diff
@@ -0,0 +1,134 @@
...
+		annotations = map[string]string{}
+	}
+
+	_, ok := annotations[annotationName]
+	if !ok {
+		// Nothing to remove
+		return false
+	}
+
+	delete(annotations, annotationName)
```
JohnStrunk, line 132: I'm trying to reconcile this w/ the setAnnotation()... in set, you call obj.SetAnnotations(), but not here. Is that because the SetAnnotations is only actually required in the case where the map was nil?

controllers/volumehandler/volumehandler.go
```diff
@@ -452,10 +490,115 @@ func (vh *VolumeHandler) ensureSnapshot(ctx context.Context, log logr.Logger,
...
+			}
+			logger.Error(copyTriggerTimeoutErr, fmt.Sprintf("Timed out after %s", utils.CopyTriggerWaitTimeout.String()))
+			vh.eventRecorder.Eventf(vh.owner, srcPVC, corev1.EventTypeWarning,
+				volsyncv1alpha1.EvRSrcPVCTimeoutWaitingForCopyTrigger, volsyncv1alpha1.EvACreateSrcCopyUsingCopyTrigger,
+				"Timed out waiting on copy trigger on src PVC %s before creating snapshot or clone",
+				utils.KindAndName(vh.client.Scheme(), srcPVC))
+			return wait, &volsyncerrors.CopyTriggerTimeoutError{
+				SourcePVC: srcPVC.GetName(),
+			}
+		}
```
JohnStrunk, line 557: Looking at this... If we have a timeout, we'll return an error, but we won't proceed w/ the replication. Then when they do finally provide the annotation, we'll proceed just like if there had been no timeout, correct?

controllers/utils/copy_trigger_annotations.go
tesshuflower: I didn't intentionally do it differently, I think in unset I figured I'm updating the map directly - let me take a look, maybe I should do SetAnnotations in both places to be consistent.

controllers/volumehandler/volumehandler.go
tesshuflower: That's correct - the timeout is really just a point at which I put up this error in the latestMoverStatus as we've been waiting for the copy-trigger to be updated for a set period of time - so similar to where we start creating events for services not coming up on time.

tesshuflower: Looking at this again, I made an update to not use the word timeout in the event or the message (I just log at .Info now).  I think it is a bit misleading to say we've "Timed out" as we're still going to keep reconciling.

tesshuflower:
@JohnStrunk I've made some small updates based on your comments - do you mind taking another look?

JohnStrunk (approved):
/lgtm

PR #1087, opened 2024-02-06
update mem limit in olm deployments of volsync

tesshuflower:

**Related issues:**


JohnStrunk (commented):
What do you think of also applying these to the Helm chart so we stay in sync?

https://github.com/backube/volsync/blob/3d91f0995ecee31be208cc4e1ae034e4bc7b8d31/helm/volsync/values.yaml#L69-L75

JohnStrunk (approved):
/lgtm

Issue #1181, opened 2024-03-24
Job keeps failing with `unable to save snapshot: snapshot is empty`


budimanjojo:
**Describe the bug**

After upgrading to v0.9.0, one of my backup job keeps failing and I don't know why.

**Steps to reproduce**

No idea yet.

**Expected behavior**

Job should run fine as the older version.

**Actual results**

**Additional context**

Here's the log of the failing pod:
```
Defaulted container "restic" out of: restic, provide-timezone (init)
Starting container
VolSync restic container version: v0.9.0+4e5e959
backup
restic 0.16.4 compiled with go1.21.8 on linux/amd64
Testing mandatory env variables
== Checking directory for content ===
== Initialize Dir =======
ID        Time                 Host        Tags        Paths
------------------------------------------------------------
1b04f8ab  2024-03-12 00:00:31  volsync                 /data
e58e09d2  2024-03-13 00:00:40  volsync                 /data
8befa5b1  2024-03-14 00:00:29  volsync                 /data
0c76ddb9  2024-03-15 00:00:42  volsync                 /data
13860d3c  2024-03-16 00:00:28  volsync                 /data
c2fb0527  2024-03-17 00:00:39  volsync                 /data
ec489a85  2024-03-18 00:00:40  volsync                 /data
a0b071d0  2024-03-19 00:00:43  volsync                 /data
6740b93d  2024-03-20 00:00:35  volsync                 /data
8a6ae86d  2024-03-21 00:00:32  volsync                 /data
------------------------------------------------------------
10 snapshots
=== Starting backup ===
/data /
using parent snapshot 8a6ae86d
Fatal: unable to save snapshot: snapshot is empty
```

I don't know what's causing the problem or how to fix it, but I'll be glad to provide more information if needed to help with this issue. Thanks before!

jalim:
I’m seeing the same issue myself if any extra info needed.  Only occurs on one of my backup jobs the remainder are all working as expected.  Only thing different is that this is the newest backup job I’ve created, not sure what version of volsync it was originally created with but likely 0.8.1

tesshuflower:
My guess is this is due to a fix I put in to ignore "lost+found" (it is new for v0.9.0).  Do you have a lost+found dir in your empty source volume?

We do have a check to see if the source is empty and skip the backup, but it will think the dir is not empty if there's a lost+found.  I think I will need to fix this empty dir check to also ignore lost+found.

budimanjojo:
[...]

Issue #1172, opened 2024-03-15
First time Replication Destinations fail in v0.8.1, succeed in v0.8.0


no1here0003:
**Describe the bug**
Creating a brand new replicationdestination in VolSync v0.8.1 fails with:
```
VolSync restic container version: unknown
restore
restic 0.16.4 compiled with go1.21.6 on linux/amd64
Testing mandatory env variables
=== Starting restore ===
Fatal: unable to open config file: Stat: The specified key does not exist.
Is there a repository at the following location?
```

v0.8.0 works.

**Steps to reproduce**
Create a new replication destination in a bucket that it has never been in before as a first time deployment in version v0.8.1, the jobs error and the pvc associated with it is stuck in Pending.

Do the same in v.0.8.0 and it works.

**Expected behavior**
It should behave as v0.8.0 did

**Actual results**
PVC is stuck in pending and VolSync jobs error with the above error message.

**Additional context**
This was a first time install of VolSync, I went straight to v0.8.1 and this happened. Downgrading to v0.8.0 worked.

no1here0003:
In case it helps, the replication endpoint is the latest release of minio.

adampetrovic:
See https://github.com/backube/volsync/discussions/1163 as well

no1here0003:
In this case I can definitely reproduce it on demand. If there's any additional information I can gather please let me know :)

tesshuflower:
There was a fix in v0.8.1 (https://github.com/backube/volsync/pull/1005) for a bug where if the auth or connection to the data store hosting the restic backup failed, the replicationdestination would succeed and not throw any errors.

If you are trying to restore a backup from a bucket that cannot authenticate or does not exist, then you will get an error.  Generally this is desired as we want to know if a restore from a backup does not succeed.

no1here0003:
Would it be possible to allow an override in the install to not error? I have a GitOps set up, so having the destination on all pvc definitions makes life a lot easier, including new ones. 

I would imagine that's what others are doing as well.

no1here0003:
Wait. Is it if a bucket doesn't exist or a path? In my case a bucket did and a path did not.

onedr0p:
> If you are trying to restore a backup from a bucket that cannot authenticate or does not exist, then you will get an error. Generally this is desired as we want to know if a restore from a backup does not succeed.

[...]

PR #1190, opened 2024-03-26
Restic restore nonexist path, ignore lost+found on backup empty dir check

tesshuflower:

- Allows a restic restore to a non-existent path - will initialize the path if it doesn't exist (as is done in a backup).  This still means backups will fail if bad credentials are supplied
- Ignore lost+found on restic backup when checking for empty source volume

Want to fix in main and then cherry-pick into release-0.9 (to be included in v0.9.1).

**Related issues:**

https://github.com/backube/volsync/issues/1172
https://github.com/backube/volsync/issues/1181


JohnStrunk (approved):
/lgtm

PR #1200, opened 2024-04-05
Tflow kubernetes v0.29.3

tesshuflower:

updates kube packages to v0.29.3

- Needed to update PVC specs to use `corev1.VolumeResourceRequirements` instead of `corev1.ResourceRequirements` - see change: https://github.com/kubernetes/kubernetes/pull/118653

- Created a new EventRecorderLogger struct which implements the new record.EventRecorderLogger interface to wrap the record.EventRecorder as the record.NewEventRecorderAdapter() function signature has changed.  See change in k8s/client-go:  https://github.com/kubernetes/kubernetes/pull/120729
  - @JohnStrunk I'm not so sure this is the best way to do this - I didn't really find a better way as controller-runtime is still returning an EventRecorder.  I've updated controller-runtime to the latest in the hopes there would be some related changes there, but it doesn't appear so.
  
- Updated controller-runtime to v0.17.2

**Related issues:**


JohnStrunk:
>   @JohnStrunk I'm not so sure this is the best way to do this - I didn't really find a better way as controller-runtime is still returning an EventRecorder.  I've updated controller-runtime to the latest in the hopes there would be some related changes there, but it doesn't appear so.

After looking through the upstream changes, I don't see a clean way to do this. They made a breaking change to the Adapter signature, and I don't see any accompanying code for consumers to use to paper over that. Also looking through the controller-runtime repo, I don't think they are going to get around to improving their event interface any time soon. This is probably the best we can do for now.

/lgtm

PR #1218, opened 2024-04-17
Fix grouping of renovate golang updates

JohnStrunk:
Renovate was supposed to be grouping all the golang updates together, but this doesn't appear to be happening (#1214, #1119). This attempts to fix the regex match so that they end up as a single PR.

**Related issues:**

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -23,7 +23,7 @@
       "groupName": "golang version",
       "groupSlug": "golang",
       "matchDepNames": [
-        "^golang$"
+        "^(golang|go)$"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


renovate.json5
```diff
@@ -91,7 +91,7 @@
   ],
   // Set max number of PRs. This is set high since we have many updates on /hold
   "prConcurrentLimit": 50,
-  "regexManagers": [
+  "customManagers": [
```
JohnStrunk, line 94: This is an update to the renovate schema

tesshuflower (approved):
/lgtm

Issue #1075, opened 2024-01-29
Allow shorter resource names for kubectl volsync and verify length not surpassing 63 char limit


clushie:
**Describe the feature you'd like to have.**
My namespace is 35 characters long, the PVC in addition also 10-15 characters.

Running
`kubectl volsync migration create --pvcname prod-demo--ji63816-terraform-module/blob --capacity 10Gi --storageclass gp2 -r md-blob`

I'll end in a situation as such:

`Service "volsync-rsync-dst-prod-demo--ji63816-terraform-module-blob-migration-dest" is invalid: metadata.name: Invalid value: "volsync-rsync-dst-prod-demo--ji63816-terraform-module-blob-migration-dest": must be no more than 63 characters`

**What is the value to the end user? (why is it a priority?)**
Long resource names will hit the limit, making it impossible to use the plugin

**How will we know we have a good solution? (acceptance criteria)**
Automatically truncate resource names to less than 63 characters, for example by using a short-id.

PR #1238, opened 2024-04-29
Kubectl volsync rsync tls

tesshuflower:

- Implements kubectl-volsync migration and replication for rsync-tls.

Migration for rsync-tls will require having stunnel installed locally.

- Also has a fix for migration to set the default replicationdestination name to something shorter, and allow optionally passing in the rdname as a parameter.

- Versioned the data structure to v2 to try to keep all params saved in the data structure identical for rsync and rsync-tls and have an option for rsync-tls enabled
- rsynctls is currently defaulted to false
- When loading v1, automatically convert it to v2 and proceed.
- it's a bit messy with all the new parameters, hopefully not too confusing.
- implemented an rsync and rsync-tls version of handlers in an attempt to separate them - will make it easier to remove rsync too if we ever get to that point.
- Testing particularly on the cli migration side isn't great, but at least was able to add new e2e tests for cli replication.

**Related issues:**
https://github.com/backube/volsync/issues/369
https://github.com/backube/volsync/issues/1075


JohnStrunk:
I really am going to review this. :cry:

Issue #1290, opened 2024-06-04
rsync-tls - fails to sync if many files in the root of the source PVC


tesshuflower:
**Describe the bug**

Rsync-tls uses globbing to get the file list to send to the destination, both these cmds can fail with `Arguments list too long` error if there are too many files (or filenames too long) for files/directories in the root directory of the source PVC.

https://github.com/backube/volsync/blob/efd604e7df311d6f7afe9c6fdbf0666c6a53d1c4/mover-rsync-tls/client.sh#L153
```
/mover-rsync-tls/client.sh: line 115: /usr/bin/ls: Argument list too long
```

Which can manifest in the client script thinking there are no files to sync

and

https://github.com/backube/volsync/blob/efd604e7df311d6f7afe9c6fdbf0666c6a53d1c4/mover-rsync-tls/client.sh#L155

Can produce the error:

```
/mover-rsync-tls/client.sh: line 117: /usr/bin/rsync: Argument list too long
```

which will result in the sync failing and repeatedly retrying.

The above log statements were taken from a v0.8.1 build.

**Steps to reproduce**

**Expected behavior**

**Actual results**

**Additional context**

PR #1291, opened 2024-06-04
Rsync tls toomanyfiles

tesshuflower:

Fixes rsync-tls issue with argments too long when many files are in the root of the source PVC 

- Made a copy of all files into /tmp/filelist.txt and then pass that as `--files-from` arg to rsync
- Needed to add `-r` flag to rsync (this flag is normally implied by `-a`, but gets turned off by default when using `--files-from`)
- added e2e tests that create 21k files 
  - used long names to reduce the number of files needed to create the issue. Hopefully this doesn't cause perf issues with the actual e2e tests in CI.  In my own env the sync does take a lot longer, but the compare pvc checks actually were surprisingly fast.
  - Also added some files on the source with subdirs and files/dirs that start with .

**Related issues:**

Fixes: https://github.com/backube/volsync/issues/1290

### DIFF
--- a/custom-scorecard-tests/config.yaml
+++ b/custom-scorecard-tests/config.yaml
@@ -197,6 +197,16 @@ stages:
         mountPath: {}
   - entrypoint:
     - volsync-custom-scorecard-tests
+    - test_rsync_tls_normal_manyfiles.yml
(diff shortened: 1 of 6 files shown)


JohnStrunk (approved):
/lgtm

PR #1316, opened 2024-06-24
fix rsync-tls copy files that start with '#'

tesshuflower:
- affected files in the root of the pvc
- for acm-12338

**Related issues:**


JohnStrunk (approved):
/lgtm

PR #1340, opened 2024-07-18
feat: Add MoverAffinity field to MoverConfig

bo0tzz:
Allow setting affinity for the mover jobs. This has multiple use cases, but what prompted it is wanting the ability to set antiAffinity between all volsync jobs so they run in sequence rather than all at once.

This seemed to me like a simple change, but I don't really have any idea what I'm doing, so all feedback is welcome.


tesshuflower:
@JohnStrunk what do you think about this enhancement - I think the code itself looks good to me.  Perhaps the greatest potential issue would be people getting into unscheduleable situations if they use this incorrectly?

I'll be away for the next week starting later today, so @bo0tzz it will probably need to wait at least until I get back ...

bo0tzz:
No worries, I have all the time in the world :slightly_smiling_face: 

> potential issue would be people getting into unscheduleable situations if they use this incorrectly

Definitely a possibility, but IMO that same risk exists with the core workload resources etc as well so I'm inclined to just chalk that up to user error.

controllers/utils/utils.go
```diff
@@ -233,6 +233,9 @@ func UpdatePodTemplateSpecFromMoverConfig(podTemplateSpec *corev1.PodTemplateSpe
 	// Security context (nil by default)
 	podTemplateSpec.Spec.SecurityContext = moverConfig.MoverSecurityContext
 
+	// Affinity
+	podTemplateSpec.Spec.Affinity = moverConfig.MoverAffinity
```
JohnStrunk, line 237: Today, I don't think we set `Affinity` anywhere else, but I'm concerned that this assignment will overwrite the field if we do someday have an automatic/default setting for this.
What about making the assignment only in the case where the user has provided a value?

```suggestion
	if moverConfig.MoverAffinity != nil {
		podTemplateSpec.Spec.Affinity = moverConfig.MoverAffinity
	}
```

JohnStrunk (commented):
I'm ok w/ this addition. There is definitely potential for users to break things, but presumably if somebody sets it, they know what they're doing.

/ok-to-test
/approve

controllers/utils/utils.go
bo0tzz: Makes a lot of sense, ty!

bo0tzz:
The logs aren't entirely clear to me - is this an issue with my code, or just a flaked test?

JohnStrunk:
It's just test flakes.

tesshuflower (approved):
/lgtm

Looks good!

PR #1341, opened 2024-07-26
feat: add support for dual stack clusters

M0NsTeRRR:
This PR allow settings service IP configuration to support dual stack clusters

No

**Related issues:**

### DIFF
--- a/helm/volsync/README.md
+++ b/helm/volsync/README.md
@@ -134,6 +134,10 @@ on the command line or via a custom `values.yaml` file.
   - Annotations to add to the operator's service account
 - `serviceAccount.name`: none
   - Override the name of the operator's ServiceAccount
+- `service.ipFamilyPolicy`: (empty)
+  - Set the ip family policy to configure dual-stack see [Configure dual-stack](https://kubernetes.io/docs/concepts/services-networking/dual-stack/#services)
+- `service.ipFamilies`: none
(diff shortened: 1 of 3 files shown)


JohnStrunk:
This PR is just exposing the IP configuration for the metrics Service, correct?
Looking through the docs on dual stack, what do you think of just setting `.spec.ipFamilyPolicy: PreferDualStack` in the template and not exposing additional configuration options?

I don't know what the specific issue is that you are addressing. My assumption is that you have a dual stack cluster, but you're only getting an IPv4 metrics endpoint.

M0NsTeRRR:
> This PR is just exposing the IP configuration for the metrics Service, correct? Looking through the docs on dual stack, what do you think of just setting `.spec.ipFamilyPolicy: PreferDualStack` in the template and not exposing additional configuration options?
In my case I want to order IP protocols but it's an opt in option.


> I don't know what the specific issue is that you are addressing. My assumption is that you have a dual stack cluster, but you're only getting an IPv4 metrics endpoint.

Yes exactly :)

JohnStrunk:
/retest

CI flakes.

tesshuflower:
The openshift clusterpools we're using to run the e2e seems to be really overloaded atm.  I'll retry this later on after some of them get cleaned up.

PR #1366, opened 2024-08-22
Allow disabling mover code with build tags

tesshuflower:
- movers are all enabled by default
- test make targets to show disabling mover code
- dockerfile would still need to be modified to exclude actual mover code, this just affects the
volsync go code for each mover
- custom-scorecard-tests also modified to test allowing for some tests not to be included if
desired.
- go build tags:
  - disable_rsync
  - disable_rsynctls
  - disable_restic
  - disable_rclone
  - disable_syncthing

**Related issues:**


Makefile
```diff
@@ -134,6 +134,14 @@ test: bundle generate lint envtest helm-lint ginkgo ## Run tests.
...
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" $(GINKGO) $(TEST_ARGS) $(TEST_PACKAGES)
 
+# Example build target showing how to run unit tests, excluding syncthing from being built entirely
+# this is not bothering to exclude syncthing during linting but should be possible
+.PHONY: test-disable-syncthing
+TEST_ARGS_DISABLE_SYNCTHING ?= --tags disable_syncthing
+test-disable-syncthing: bundle generate lint envtest helm-lint ginkgo ## Run tests.
+	-rm -f cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" $(GINKGO) $(TEST_ARGS) $(TEST_ARGS_DISABLE_SYNCTHING) $(TEST_PACKAGES)
+
```
JohnStrunk, line 144: I'm wondering whether we can make this more general. I know it's just an example, but here's what I'm thinking...

Usage:
```console
$ TAGS=disable_syncthing make <target(s)>
```

We could then have at the top of the Makefile something that picks up whether TAGS is defined and sets secondary variables for build flags for go and for the container build arg...
```makefile
GOBUILDTAGS:=
CONTANIERBUILDTAGS:=
ifdef TAGS
    GOBUILDTAGS:=-tags $TAGS
    CONTAINERBUILDTAGS:=--build-arg "tags_arg=$TAGS"
endif
```
... then we just always add GOBUILDTAGS, CONTAINERBUILDTAGS at the appropriate places

The syntax probably has errors, but it's the general idea... thoughts?

tesshuflower: I like this idea, one thing is if someone does build with a mover disabled, this won't disable actually packaging the mover itself (such as syncthing for example) - it would just disable the syncthing mover code in the volsync manager binary.  Do you think it's worth looking at disabling the mover packaging in the makefile too?  I guess I was hoping the example make targets are showing some work might be needed, but here's a reference for later if we get there.

Another current limitation is unit tests won't all run well if we disable certain movers - rsync is required for the controller tests (wonder if I should change these to use rsync-tls actually), and we have some rclone mover tests in controllers/ as well so currently there's some setup in suite_test.go that requires rclone.

tesshuflower:
@JohnStrunk I've updated the Makefile, I think it works pretty well with the changes you suggested.

I also made a change to update the err we put in the status conditions if the mover isn't found to list what movers are enabled - so you get something like this:

```
status:
  conditions:
  - lastTransitionTime: "2024-08-29T17:52:07Z"
    message: 'a replication method must be specified - enabled movers: [rclone rsync
      rsync-tls]'
    reason: Error
    status: "False"
    type: Synchronizing
```

This change is just with the last commit in the PR, so we can roll it back if you think it's overkill.

PR #1378, opened 2024-09-05
Update restic to v0.17.1

tesshuflower:
- Updates restic to v0.17.0 *edit* - updates to v0.17.1
- Make sure new AWS env vars are passed through

**Note - not including any changes atm for the following**

 There are some new features which we may or may not want to expose in restic, but will do separately from this PR and may create enhancement issues for some of them - recording them here for future reference:

Backup features:
- `--skip-if-unchanged` flag - omits snapshot creation if no change from parent snapshot

Restore features:
- `--delete` flag - removes files on restore that don't exist in the backup - would allow for repeatedly running restores to a PVC
- `--overwrite` flag - overwrite behavior on restore
- Restore also has new options to include/exclude patterns from files `--include-file`, `--exclude-file`, `--iinclude-file` and `--iexclude-file`  (note `--include` and `--exclude` flags already existed and these aren't currently exposed)...would be easier to expose those than the -file ones.

Misc:
- allow repos with empty password `--insecure-no-password` - not sure if this is worth supporting atm
- snapshot size output enhancement: should we save any of this in the mover log status?

**Related issues:**


JohnStrunk (approved):
/lgtm

PR #1383, opened 2024-09-10
Restic restore --delete flag

tesshuflower:

Allows passing the `--delete` flag to restic restores.

New field in the `replicationdestination` spec:

`spec.restic.delete`

Do we want to name the field in the api spec something different? I used `delete` to match the restic restore flag. 

**Related issues:**

### DIFF
--- a/controllers/mover/restic/builder.go
+++ b/controllers/mover/restic/builder.go
@@ -196,6 +196,7 @@ func (rb *Builder) FromDestination(client client.Client, logger logr.Logger,
 		privileged:            privileged,
 		restoreAsOf:           destination.Spec.Restic.RestoreAsOf,
 		previous:              destination.Spec.Restic.Previous,
+		deleteFilesOnRestore:  destination.Spec.Restic.Delete,
 		latestMoverStatus:     destination.Status.LatestMoverStatus,
 		moverConfig:           destination.Spec.Restic.MoverConfig,
 	}, nil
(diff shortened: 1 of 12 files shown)


controllers/mover/restic/builder.go
```diff
@@ -196,6 +196,7 @@ func (rb *Builder) FromDestination(client client.Client, logger logr.Logger,
 		privileged:            privileged,
 		restoreAsOf:           destination.Spec.Restic.RestoreAsOf,
 		previous:              destination.Spec.Restic.Previous,
+		deleteFilesOnRestore:  destination.Spec.Restic.Delete,
```
JohnStrunk, line 199: Not a fan of this name as it can easily be read as "we're going to delete all files in the PVC before we start the restore." `allowDeleteOnRestore`?

tesshuflower: I was struggling to come up with a good name for this...

I've been waffling back and forth about this api name and the name I use in the builder (ideally I'd like to use the same name in both places, but think I ended up with a different name to indicate this was a restore only parameter).  

I do think I like `enableFileDeletion` for the API -  So what do you think about `enableFileDeletionOnRestore` for the builder.go? @JohnStrunk

tesshuflower:
@JohnStrunk I've made updates to those var names, could you take another look?

JohnStrunk (approved):
/lgtm

PR #1389, opened 2024-09-15
chore: don't log if privileged-movers set to false

michaelbeaumont:
Gives users the option to reduce noise in the logs


controllers/utils/namespace.go
```diff
@@ -49,10 +49,11 @@ func PrivilegedMoversOk(ctx context.Context, cl client.Client, logger logr.Logge
 			"namespace", namespace, "Annotation", volsyncv1alpha1.PrivilegedMoversNamespaceAnnotation,
 			"Annotation value", privilegedMoverOkVal)
 		return true, nil
+	} else if strings.ToLower(privilegedMoverOkVal) != "false" {
+		logger.Info("Namespace does not indicate volsync privileged movers are allowed, will assume privileged movers "+
+			"are not allowed. Annotation missing or not set to 'true' or 'false'", "namespace", namespace,
+			"Annotation", volsyncv1alpha1.PrivilegedMoversNamespaceAnnotation)
```
tesshuflower, line 55: @michaelbeaumont thanks for this, every time I see that log entry I think I should remove it, but then forget :) Since privileged movers are used by default for a while now, that I'd actually prefer to remove the logging that privileged=false is assumed entirely.  We can keep the log line if someone enables privileged movers on the namespace.

```suggestion
```

PR #1395, opened 2024-09-25
use kube 1.31.0 for tests

tesshuflower:
- adds 1.31.0 to test matrix in operator.yml
- updates unit tests to use 1.31.0 by default
- sets up kind env with 1.31.0 by default

**Related issues:**


JohnStrunk (approved):
/lgtm

PR #1396, opened 2024-09-25
Controller runtime and tools update

tesshuflower:

- updates controller-runtime to v0.19.0 (commit from renovatebot)
- This updates most k8s libs to v0.31.0
- updates controller tools to v0.16.3 (old controller-gen doesn't work anymore with these updates), and generate manifests etc

- running make manifests does some optimization to consolidate the roles so roles.yml looks quite a bit different, but appears to be ok as far as I can tell.

**Related issues:**


JohnStrunk (approved):
/lgtm

PR #1412, opened 2024-10-08
Revert "Update restic to v0.17.1"

tesshuflower:
This reverts commit c3e546894d4170aa1e5fb1d552306915bd047020.

Rolls back to restic v0.17.0 for the moment.

- Had issues on OpenShift with v0.17.1, due to this fix in restic: https://github.com/restic/restic/pull/4958

- Essentially now when running in privileged mode, restic will fail if restoring extended attributes fails.  This will fail on OpenShift unless the namespace annotations for `openshift.io/sa.scc.mcs` match on the replicationdestination namespace.  This is because OpenShift will not allow restoring these settings (unless we brute force allow privileged escalation and allow the pod to run in privileged mode, instead of just running as root).
- Other extended attributes could also cause the restore to fail, such as security.* attributes.  To restore those, we'd need to grant the privileged mover pod `SYS_ADMIN` capabilities.

- For the moment we will roll back to v0.17.0 to maintain the current behavior, but may look into seeing if restic will allow flags to skip backing up/restoring specific extended attributes.

Will want to cherry-pick this into release-0.11

**Related issues:**

https://issues.redhat.com/browse/ACM-14701


JohnStrunk (approved):
/lgtm

PR #1420, opened 2024-10-11
update ensure_initialized in restic mover script

tesshuflower:

Updates the ensure_initialized function in the restic mover script.

I came across this info in the restic docs:  https://restic.readthedocs.io/en/stable/075_scripting.html#check-if-a-repository-is-already-initialized

Essentially it recommends using `restic cat config` in order to determine if a repo has been initialized.

A few things:
- It looks like there are some new return codes that will maybe help us show a better error if we can't connect to the repo
- Ideally I'd like to remove the check for return code 1 and return an error here (looks like generally rc=10 should mean the repo doesn't exist, and at that point we could run restic init).  However in cases like minio, if the bucket does not exist, we still get rc=1....so I left the previous behaviour in there for this scenario.
- I check for rc=12 which won't be there until we use restic v0.17.1 or later, but when we do get it, we can have a proper err message about bad password.

Generally this may not be needed as restic init is still supposed to fail if the repo has been previously initialized, but this moves more towards the recommended way for restic.

Also adds an e2e test for the gitops scenario of restoring from a non-existent bucket/repo.

**Related issues:**

### DIFF
--- a/controllers/mover/restic/logfilter.go
+++ b/controllers/mover/restic/logfilter.go
@@ -37,7 +37,10 @@ var resticRegex = regexp.MustCompile(
 		`^\s*([aA]dded to the repository)|` +
 		`^\s*([sS]uccessfully)|` +
 		`(RESTORE_OPTIONS)|` +
-		`^\s*(Restic completed in)`)
+		`([iI]nitialize [dD]ir)|` +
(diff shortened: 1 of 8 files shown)


JohnStrunk (approved):
/lgtm

Issue #1454, opened 2024-11-13
`ReconciliationSource` resources using the `restic` mover will silently fail under certain conditions.


benfiola:
**Describe the bug**
It appears that that the `restic` has a [code path](https://github.com/backube/volsync/blob/b36e2df68dd20b3ddd126c952f83c4f328a58d84/controllers/mover/restic/mover.go#L234) where errors are inadvertently suppressed.  This code path includes cases where a `ReplicationSource` might have invalid data - which was the case when I accidentally left `spec.restic.copyMethod` unset.  Ultimately, I had to clone the repo locally and run the controller with a debugger to determine why this resource was failing to reconcile.

**Steps to reproduce**
Create a `ReplicationSource` using `restic` but without setting `spec.restic.copyMethod`.  

**Expected behavior**
The resource should be reconciled (e.g., the data moving jobs should be created).

**Actual results**
The resource is updated to the `Transitioning` state and then nothing happens.

PR #1455, opened 2024-11-13
restic mover should return error if unable to 'EnsurePVCFromSrc'

benfiola:
It appears that the restic mover will suppress errors when `EnsurePVCFromSource` is called.  If the `ReplicationSource` has incorrect values (- in my case, I had forgotten the `spec.restic.copyMethod` field) - the resource will silently fail to re-reconcile.

None that I can immediately see.

**Related issues:**
This PR addresses the following [issue](https://github.com/backube/volsync/issues/1454)

(Aside from this hiccup caused by user error - thank you for the excellent controller!  It's exactly what I've been looking for and works great.)

### DIFF
--- a/controllers/mover/rclone/mover.go
+++ b/controllers/mover/rclone/mover.go
@@ -187,7 +187,7 @@ func (m *Mover) ensureSourcePVC(ctx context.Context) (*corev1.PersistentVolumeCl
 			return pvc, nil
 		}
 	}
-	return pvc, nil
+	return pvc, err
 }
 
 // this is so far is common to rclone & restic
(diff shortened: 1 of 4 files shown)


tesshuflower:
@benfiola thanks for this!

I'm the one who introduced this bug :grimacing: I've just pushed another commit that makes the same fix for other movers that have the same issue.

benfiola:
Sounds good!  Closing this one im favor of your changes!

tesshuflower:
@benfiola Oops, I should've been clearer - the commit I pushed was to your branch - so still need to get this PR merged (contains your change, plus my additional change to do the same for other movers).

benfiola:
OH, my bad!  😅 I’ll stop touching the buttons.

PR #1466, opened 2024-11-23
Sync only the target filesystem at the end of mover tasks

evandeaubl:
This PR fixes a bug I observed in a cluster where I had many rsync-tls privileged volsync jobs running against Longhorn volumes, and at least one was doing a large transfer. After a short time, the load average on the node with the large transfer destination mover would shoot really high (around 15), and the node started having problems at the kubelet level (pods failing to start with CreateContainerError was the easiest to observe condition)

Digging deeper, I saw that the other destination mover pods on that node (not related to the large transfer) were hanging at the ends of their transfers, when they should have been exiting and recycled by the operator.

Comparing the logs to the script showed that the only thing that could have been hanging was the `sync` command at the end of the script. Execing into the container and running a manual sync hanged in the same way, confirming that was the issue. If I instead limited the `sync` to just the target filesystem using `sync -f /data`, the hang did not occur. Building a custom rsync-tls mover with the `sync` command limited to the one filesystem resolved this issue.

My understanding of how sync works in containers makes me think the general `sync` was trying to sync much more than just what was in the container itself, and combined with an active large transfer was causing things to bind up.

I am only running rsync-tls at the moment, but there were `sync` commands in the rclone and rsync-ssh movers that could exhibit the same problem that I fixed as well, but did not test.

**Related issues:**
None

### DIFF
--- a/mover-rclone/active.sh
+++ b/mover-rclone/active.sh
@@ -87,5 +87,5 @@ destination)
     error 1 "unknown value for DIRECTION: ${DIRECTION}"
     ;;
 esac
-sync
+sync -f ${RCLONE_DEST_PATH}
 echo "Rclone completed in $(( SECONDS - START_TIME ))s"
--- a/mover-rclone/active.sh
+++ b/mover-rclone/active.sh
@@ -87,5 +87,5 @@ destination)
     error 1 "unknown value for DIRECTION: ${DIRECTION}"
     ;;
 esac
-sync -f ${RCLONE_DEST_PATH}
+sync -f "${MOUNT_PATH}"
 echo "Rclone completed in $(( SECONDS - START_TIME ))s"
--- a/mover-restic/entry.sh
+++ b/mover-restic/entry.sh
@@ -376,7 +376,7 @@ for op in "$@"; do
         "restore")
             ensure_initialized
             do_restore
-            sync
+            sync -f "${DATA_DIR}"
             ;;
         *)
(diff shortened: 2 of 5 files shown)


tesshuflower:
@evandeaubl thanks for this great analysis and your PR! I do feel like we've heard about the occasional problem with running sync, hope this will address most of them.  I agree it makes sense to only run the sync against the mounted volume.

I've just pushed a small change with updates to some of the scripts.

evandeaubl:
@tesshuflower Thanks for the script cleanup, and catching that one `sync` I missed. Easy to miss finding a `sync` in a project named volsync. 😄

Also, I've been testing this change over the past few days on my homelab cluster doing some pretty large transfers under the same conditions as described. The change has been working great, and hasn't caused any regressions that I've observed.

tesshuflower:
@evandeaubl Sorry for the delay on this one - there were some issues with automated tests that were breaking all the e2e tests from github actions.  Unfortunately this means we'll have to rebase this PR to pickup the latest updates in main. 

I can do this, but would require force-pushing and that will re-write your history in your local fork/main.  You can either rebase yourself, or let me know if you don't mind and I can do it.  Alternatively I can also pull in your changes into a new PR if you prefer - just let me know!

evandeaubl:
@tesshuflower No problem with you rebasing it. It doesn't disrupt anything on my end, and I don't think I'd have spare cycles to look at this until towards the end of the week. Thank you!

PR #1478, opened 2024-12-06
Rsync e2e usepam no

tesshuflower:

UsePam no for sshd_config
    
e2e running on ubuntu 24.04 (not sure of exact cause) is having issues with sshd allowing root access which is required for rsync ssh.  Seems to be related to PAM.

Very strange and unsure of exact cause, but seems to be recreatable with kind running on an ubuntu 24.04 host.  Something in PAM is simply failing and not allowing root to login.

**Related issues:**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ COPY go.sum go.sum
 RUN go mod download
 
 # Copy the go source
-COPY *.go .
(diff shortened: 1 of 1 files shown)


JohnStrunk (approved):
/lgtm

PR #1490, opened 2024-12-17
helm update - put kube-rbac-proxy img in values

tesshuflower:

Put default value for kube-rbac-proxy in helm values.yaml, will allow it to be overriden more easily if a user wants.

**Related issues:**

### DIFF
--- a/helm/volsync/values.yaml
+++ b/helm/volsync/values.yaml
@@ -33,6 +33,8 @@ syncthing:
   # Overrides the image tag whose default is the chart appVersion.
   tag: ""
   image: ""
+kube-rbac-proxy:
+  image: quay.io/brancz/kube-rbac-proxy:v0.18.1
 
 manageCRDs: true
 
--- a/helm/volsync/values.yaml
+++ b/helm/volsync/values.yaml
@@ -34,7 +34,9 @@ syncthing:
   tag: ""
   image: ""
 kube-rbac-proxy:
-  image: quay.io/brancz/kube-rbac-proxy:v0.18.1
+  repository: quay.io/brancz/kube-rbac-proxy
+  tag: "v0.18.1"
+  image: ""
 
 manageCRDs: true
 
--- a/helm/volsync/templates/deployment-controller.yaml
+++ b/helm/volsync/templates/deployment-controller.yaml
@@ -52,7 +52,7 @@ spec:
               drop:
                 - "ALL"
             readOnlyRootFilesystem: true
-          image: quay.io/brancz/kube-rbac-proxy:v0.18.1
+          image: "{{ include "container-image" (list . (index .Values "kube-rbac-proxy") ) }}"
           args:
             - --secure-listen-address=0.0.0.0:8443
             - --upstream=http://127.0.0.1:8080/


helm/volsync/values.yaml
```diff
@@ -33,6 +33,8 @@ syncthing:
   # Overrides the image tag whose default is the chart appVersion.
   tag: ""
   image: ""
+kube-rbac-proxy:
+  image: quay.io/brancz/kube-rbac-proxy:v0.18.1
```
JohnStrunk, line 37: Can we follow the `repository`, `tag`, `image` pattern that we use for the other images or is there something special about the rbac proxy?

tesshuflower: I guess I thought it would simply make the renovate replacement easier, as we would normally replace the entire image path here: https://github.com/backube/volsync/blob/main/config/default/manager_auth_proxy_patch.yaml#L35

If we ever didn't set tag, it also means we'd default to the volsync appVersion, but that probably isn't a real concern.

But...we could also do:

```
kube-rbac-proxy:
  repository: quay.io/brancz/kube-rbac-proxy
  tag: v0.18.1
  image: ""
```

Maybe the best question is:  Can renovate automatically update this and which format is easiest?  I didn't see a specific rule for kube-rbac-proxy but we do have something for docker images in yamls...which maybe would work on the values.yaml as well?

https://github.com/backube/volsync/blob/main/renovate.json5#L152-L161

vistasunil:
We need to Pull request to be merged to use Volsync next week for our migration. Is it possible to merge this pull request this week?

helm/volsync/values.yaml
JohnStrunk: We can customize renovate to handle the separate tag field.

tesshuflower: Gotcha, will update to use repository & tags then 👍

JohnStrunk (approved):
/lgtm

vistasunil:
@JohnStrunk I see this PR is already merged yesterday but cant see any release out of it. Though one release 0.11.1 is created but again these image changes are not there. So, can you please suggest how these changes will be available for us to use? Will there be any separate release or its already there?

tesshuflower:
@vistasunil These changes are only in main, so will only make it into the next release (which will be 0.12.0), most likely in February sometime.

There aren't any significant changes in the helm charts themselves, so you could try using the helm charts in main - but you would need to update the values to set the image labels to use 0.11.0 (or "latest" if you want the latest available builds).

We can consider releasing a v0.12.0-rc.1 build (this would be a pre-release build) which would include the latest helm charts if that would help

vistasunil:
@tesshuflower Yes.. that will help. Please release  v0.12.0-rc.1 this week if possible. Because we need to deploy and use it from next week itself.

tesshuflower:
pre-release v0.12.0-rc.1 is now released - note this is still a pre-release and you may need to use the ```--devel``` flag with helm install to deploy it.

vistasunil:
@tesshuflower thank you :)

vistasunil:
@tesshuflower I can see that release 0.12 is still not in place.  Can you please share an update on this?

PR #1540, opened 2025-02-03
Mover pull secrets

tesshuflower:

When imagePullSecrets are set in the volsync helm charts, will add cli command arg --mover-image-pull-secrets to the volsync controller.

When mover-image-pull-secrets are set, when reconciling a `replicationsource` or `replicationdestination` service account, the secret(s) will be copied from the volsync controller namespace (normally `volsync-system`) to the mover namespace and added to the service account created by volsync.

- In the mover namespace, only 1 secret will be created for each original secret in the volsync-system namespace.  These will be named based on a common prefix "volsync-pull-" plus a crc32 hash of the original secret name.
- If there are multiple volsync CRs in the namespace, they will all use the same secrets, but will add their own owner reference to them. This way the last volsync CR to be cleaned up in a namespace should invoke garbage collection to remove the secrets themselves.
- If the secret doesn't exist in the volsync controller namespace, it will just be logged, but will not throw an error.  This is meant to be consistent with pull secrets in that if they do not exist, there will be no error (assuming you are able to pull the image via one of the secrets)

**Related issues:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- imagePullSecrets in helm charts will be copied from volsync controller
+  namespace to the mover namespace so mover jobs can use them
 - Updated release to build on golang 1.23
 - Syncthing updated to v1.29.2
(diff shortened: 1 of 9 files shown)


JohnStrunk (approved):
/lgtm

Issue #1430, opened 2024-10-22
Volsync does not start in cluster with PodSecurityAdmission set to restricted


davralin:
**Describe the bug**
VolSync fails to start in a cluster with Pod Security Configuration enforced to restricted, due to missing SecCompProfile.

**Steps to reproduce**
Enforce pod security configuration to restricted in your cluster.
Deploy volsync

**Expected behavior**
Volsync is actually deployed

**Actual results**
````
Error creating: pods "volsync-554964465c-lw8t4" is forbidden:
violates PodSecurity "restricted:latest":
seccompProfile (pod or containers "kube-rbac-proxy", "manager" must set securityContext.seccompProfile.type
to "RuntimeDefault" or "Localhost")
````

**Additional context**

tesshuflower:
You may need to set `podSecurityContext.seccompProfile.type` in your values.yaml

https://github.com/backube/volsync/blob/main/helm/volsync/values.yaml#L66

davralin:
Right... not sure how I missed that...
Anyway;
````yaml
    podSecurityContext:
      runAsNonRoot: true
      seccompProfile:
        type: RuntimeDefault
````
Leads to:
````
pod/volsync-5cc98f9b59-plvkf
Error: container has runAsNonRoot and image will run as root
(pod: "volsync-5cc98f9b59-plvkf_volsync(d7b7b3c9-8959-48ca-b8b3-b8a226c21c12)", container: manager)
````yaml

So, new thing failed - but probably perhaps somewhat related anyway.

tesshuflower:
I think it sounds like you also will need to set `podSecurityContext.runAsUser` in your system.  By default we normally set this to 65534 if you don't specify a podSecurityContext.

davralin:
So, this is the conclussion:

````yaml
    podSecurityContext:
        runAsUser: 1000
        runAsGroup: 1000
        fsGroup: 1000
        fsGroupChangePolicy: "OnRootMismatch"
        runAsNonRoot: true
        seccompProfile:
          type: RuntimeDefault
````

This will allow volsync to work under restricted.

In order to do backup/restores:

````yaml
    moverSecurityContext:
      runAsUser: 1000
      runAsGroup: 1000
      fsGroup: 1000
      fsGroupChangePolicy: "OnRootMismatch"
      runAsNonRoot: true
      seccompProfile:
        type: RuntimeDefault
````

Maybe some of these values should be the defaults?

[...]

PR #1564, opened 2025-02-22
feat(seccompProfile): enable seccompProfile by default

davralin:
Saw that the setting was commented out in the controller, this is an attempt at setting it as default.

I'm thinking that people either want to replace the entire podsecurityContext-value, and not the spesific seccompProfile-value.

Really don't know what I'm doing here, just wanted to try contributing.

**Related issues:**
#1430

### DIFF
--- a/helm/volsync/templates/deployment-controller.yaml
+++ b/helm/volsync/templates/deployment-controller.yaml
@@ -39,8 +39,8 @@ spec:
         {{- if not (.Capabilities.APIVersions.Has "security.openshift.io/v1/SecurityContextConstraints") }}
         runAsUser: 65534
         {{- end }}
-        # seccompProfile:
-        #   type: RuntimeDefault
+        seccompProfile:
+          type: RuntimeDefault
       {{- else }}
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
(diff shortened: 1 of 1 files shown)


tesshuflower:
thanks @davralin - I pushed a small update only to do the seccompprofile for non-openshift, as for openshift we'll let it set that for us.  And I thought I might as well specify the runAsGroup while we're at it.

davralin:
> thanks @davralin - I pushed a small update only to do the seccompprofile for non-openshift, as for openshift we'll let it set that for us.  And I thought I might as well specify the runAsGroup while we're at it.

Cool, not familiar with openshift.

tesshuflower:
> > thanks @davralin - I pushed a small update only to do the seccompprofile for non-openshift, as for openshift we'll let it set that for us.  And I thought I might as well specify the runAsGroup while we're at it.
> 
> Cool, not familiar with openshift.

Note, John's pointed out to me that I was wrong about OpenShift - setting seccompProfile only caused an issue with really old versions (< 4.11) which we no longer support.  So I'm creating a new PR to update for OpenShift as well: https://github.com/backube/volsync/pull/1569

PR #1569, opened 2025-02-25
Kind enforce restricted and seccomprofile ocp

tesshuflower:
- Updates to set seccompProfile to RuntimeDefault for OpenShift (previously we didn't set for Openshift, but this was only an issue with OpenShift <= 4.11 which is no longer supported).
- Updates the setup-kind-cluster.sh to enforce restricted by default
- some updates to e2e tests that need to run privileged but didn't set the label on the ns to allow privileged pods
- updates minio to an updated chart version that sets seccompProfile by default to RuntimeDefault

**Related issues:**


JohnStrunk (approved):
/lgtm

PR #1609, opened 2025-03-28
Restic v0.18.0

tesshuflower:

- Update restic to v0.18.0

Also updated `mover-restic/entry.sh` to use the`--include-xattr user.*` flag on restic restore to restrict restore of xattrs to only the user namespace.

- see issue: https://github.com/restic/restic/issues/5089

    This issue occured only in OpenShift when using restic in
    privileged mode. With restic v0.17.1 and later, restoring xattrs
    would cause the restore to fail if UID=0 if there were any
    issues restoring xattrs.

    When using privileged, we run as uid 0 and on some kube clusters
    such as OpenShift, we were unable to restore things like
    security.selinux settings. This had never worked, but previous
    verions of restic woudl not consider it an error.

    Workaround right now is to stop trying to restore security,
    system and trusted namespaced xattrs.

    Easiest just to use the --include-xattrs flag to specify that
    we only want to attempt to restore user.* xattrs.

- This version of restic also introduced a feature which breaks our automatic check if the dir is initialized:  https://github.com/restic/restic/issues/5081

  We do a `restic cat config` to determine if the dir can be reached and if it's initialized, but now this may start retrying for 15 minutes.  Note that we previously used `restic snapshots` for this check, and it turns out this also has retries now.  What this means is in our tests that use minio, when the base bucket does not exist, it will keep retrying.  Added a timeout command to timeout the `restic cat config` after 10s and then after that we proceed as usual.

**Related issues:**


JohnStrunk (approved):
/lgtm

Issue #1575, opened 2025-03-04
Cache Volume Name Conflict


manikandanmr:
### Context:
When creating a `ReplicationSource` and a `ReplicationDestination` with the same name (for example, "demo"), VolSync generates associated volumes with a naming pattern that includes a shared cache volume. For instance:

**ReplicationSource "demo" creates:**

- `volsync-demo-src`
- `volsync-demo-cache`

**ReplicationDestination "demo" creates:**

- `volsync-demo-dest`
- `volsync-demo-cache`

### Problem:
The cache volume (`volsync-demo-cache`) is created by both objects. Since Kubernetes enforces unique ownership on PVCs via owner references, only one object can own this cache PVC. As a result, when both a replication source and destination share the same name, one of them ends up "locking" the cache volume. This leads to errors during reconciliation, such as:

```
ERROR: volsync-demo-cache is already owned by another ReplicationDestination controller
```

### Impact:

- The replication source process fails to backup the volume because the cache volume is already claimed by one of the objects.

### Recommendation:

- Investigate whether naming conventions can be adjusted so that the cache volume names are unique between a replication source and a replication destination.

tesshuflower:
We will try to fix this one (restic only).

Note that it's a little more complicated as we will probably need to cleanup old cache volumes created with the old name for existing replicationsource/replicationdestinations.

I won't get to this for a bit, but have put it on my list of things to look at for next release (probably v0.13 timeframe) - @manikandanmr in the meantime you'd need to create your replicationsource and destination with different names.

PR #1611, opened 2025-04-01
fix restic cache pvc name collisions for src & dst

tesshuflower:
Fixes: https://github.com/backube/volsync/issues/1575

- Will check for old cache pvc (with the previous name) and remove it if it exists, and then create/reconcile a cache pvc with the new name

**Related issues:**

https://github.com/backube/volsync/issues/1575


JohnStrunk (approved):
/lgtm

PR #1620, opened 2025-04-10
Golangv4 plugin upgrade operator sdk upgrade to v1.33.0

tesshuflower:

- migrates to use operator-sdk v1.33.0
- includes updates to move controller code from /controllers to /internal/controller
- yaml updates due to new kubebuilder

- We were embedding the mover_scc.yaml to be able to keep it up-to-date and have the same file in our /config yamls - but this file existed in /config/openshift - moved it to /internal/controller/openshift so it could still be embedded, and then made a symlink in /config/openshift/mover_scc.yaml -> ../../internal/controller/openshift/mover_scc.yaml

**Related issues:**


JohnStrunk (approved):
I'll admit, I don't think I followed the changes to the metrics port.
/lgtm

tesshuflower:
> I'll admit, I don't think I followed the changes to the metrics port. /lgtm

Thanks John!  Some of it was due to the kubebuilder migration I had run...but I ended up having to add our metrics patch stuff back in - I think in the end it all ends up looking pretty much the same - but if I move away from the kube-rbac-proxy in the end, then I think I'll have to modify the kustomize script that updates for metrics

PR #1629, opened 2025-04-17
fix gen-files check for ffwd branch

tesshuflower:

release-0.13 br has a check to make sure the custom scorecard tests are setup to use "release-0.13" tag instead of "latest"...but if we're ffwding main -> release-0.13 then this won't work.  

**Related issues:**


JohnStrunk (approved):
/lgtm

PR #1649, opened 2025-04-25
Dockerfile.rhtap add back fix to not strip symbols

tesshuflower:
- add back steps we had in the upstream so symbols are not stripped for rclone & restic (so we can verify crypto libs with nm)

**Related issues:**


JohnStrunk (approved):
/lgtm

PR #1656, opened 2025-05-01
update mover-restic/minio-go/go.mod

tesshuflower:

- update to replace `go 1.22` with `go 1.22.0` in the copy of mover-restic/minio-go/go.mod
- cachi2 tool used in downstream builds fails in a prefetch stage as it does not find a toolchain for 1.22 (needs the .z specified).  konflux has a prefetch stage to pickup golang dependencies that runs cachi2.

**Related issues:**

### DIFF
--- a/mover-restic/update-restic-to.sh
+++ b/mover-restic/update-restic-to.sh
@@ -57,7 +57,6 @@ log "Updating minio-go source to tag: $MINIO_GO_TAG"
 update_repo_to "$MINIO_GO_REPOSITORY" "$MINIO_GO_TAG" minio-go
 
 
-
 ############################################################
 ## Patch minio-go
 ############################################################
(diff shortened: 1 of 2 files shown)


mover-restic/update-restic-to.sh
```diff
@@ -57,14 +57,15 @@ log "Updating minio-go source to tag: $MINIO_GO_TAG"
...
-
 ############################################################
 ## Patch minio-go
 ############################################################
 
 cd minio-go
 # Remove sha256-simd library
 find . -name '*.go' -exec sed -ri 's|github.com/minio/sha256-simd|crypto/sha256|' {} \;
+# Downstream builder needs to pre-fetch files with cachi2 and cannot handle go 1.22 (needs go 1.22.0 for the toolchain)
+sed -i 's|go 1.22$|go 1.22.0|' go.mod
```
JohnStrunk, line 68: My concern here is that this is going to break when they update to 1.23, etc. What about a check/patch to ensure a `.z` exists. Could adding an arbitrary `.0` to an X.Y work or are there really only builders for specific Z?

tesshuflower: I'm hoping this will be temporary, and only for go 1.22.  I'm not sure I understand the whole toolchain thing that well, but from what I could tell in the konflux discussions about cachi2 was this is a go thing where they pull in the toolchain and go 1.22 is not actually correct (so they won't be modifying the cachi tool to do this .z stuff for us).

I think when moving to go 1.23 you're actually forced to add the .z and/or have a toolchain statement, and the latest in minio-go already does this actually.  So this workaround should only needed until restic points to a version of minio-go that uses 1.23.

Issue #1710, opened 2025-07-18
Restic scheduled replicationsource fails with v0.13.0


michaelbeaumont:
**Describe the bug**
The backup job fails with `signal terminated received`:

```
Starting container
VolSync restic container version: v0.13.0+34b8e30
backup
restic 0.18.0 compiled with go1.23.11 on linux/amd64
Testing mandatory env variables
== Checking directory for content ===
=== Check for dir initialized ===
signal terminated received, cleaning up
unable to create lock in backend: context canceled
ERROR: failure checking existence of repository
```

**Steps to reproduce**
Upgrade to v0.13.0 and run a restic backup job

**Expected behavior**
Job succeeds. Downgrading to v0.12.1 makes it work again

**Actual results**
It fails continually

**Additional context**
I don't see any events on the Pod or the Job that would indicate why the Pod is being terminated. It exits with code `3`.
I'm using Backblaze via the S3 API

tesshuflower:
The main change with VolSync v0.13.0 with restic is that restic has been updated from `v0.17.0` to `v0.18.0`.

I haven't been able to recreate this so far - tried the following (these are all using minio as the s3 storage location):

- created a backup using volsync v0.13.0 (creates new restic repository)
- created a subsequent backup using v0.13.0 (re-uses existing restic repository above)
- created a backup using volsync v0.12.0 (creates a new restic repository)
- created a backup using volsync v0.13.0 with the restic repository created by volsync v0.12.0

My thinking was that this was most likely some sort of migration issue using the repo used previously in restic v0.17.0 to restic v0.18.0.

However maybe it's more complicated than that.

Some details - from the log you provided, VolSync is using restic to ensure the repository exists before continuing and creating a new backup.

That is, it will issue the command:

```
restic cat config
```

And this is returning:
```
unable to create lock in backend: context canceled
```

Would it be possible for you to point a restic v0.18.0 at your repository and issue the same command? 

Another interesting thing to try would be to see if v0.18.0 works with a new empty location (or volsync backing up to a different path) - forcing it to create a new restic repository.  Essentially this could be some issue with the repository itself (created in an older version of restic) and restic v0.18.0, or perhaps could be an issue with restic v0.18.0 and your particular storage.

michaelbeaumont:
I'll try that. I'll add that my backend is backblaze

michaelbeaumont:
Unfortunately? `restic cat config` is successful with v0.18

tesshuflower:
[...]

PR #1714, opened 2025-07-22
increase timeout on restic cat config

tesshuflower:

When moving to restic v0.18.0 we updated the check initialized as the recommended approach of `restic cat config` now  could hang retrying for up to 15 minutes if the repo wasn't accessble.

We then timeout after 10s - but looks like for some users this may take significantly longer - up to 40s mentioned in the issue below.

Long term maybe we can find a better solution - but would be nice to increase the timeout for now, at least for a v0.13.1

**Related issues:**

https://github.com/backube/volsync/issues/1710

### DIFF
--- a/mover-restic/entry.sh
+++ b/mover-restic/entry.sh
@@ -95,7 +95,7 @@ function ensure_initialized {
     set +e  # Don't exit on command failure
 
     outfile=$(mktemp -q)
-    timeout 10s "${RESTIC[@]}" cat config > /dev/null 2>"$outfile"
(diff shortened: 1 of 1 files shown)


michaelbeaumont:
@tesshuflower is it just the e2e we're waiting on here? I don't wanna spam / retest

tesshuflower:
@michaelbeaumont you can leave it right now - I'll need to wait for next week when John is back from vacation before I deliver anything, and we may look at breaking out of the timeout early if we see the retries ...

tesshuflower:
/close

Closing in favour of: https://github.com/backube/volsync/pull/1744

Issue #671, opened 2023-03-07
Support SFTP destinations in Restic


csnyder616:
**Describe the feature you'd like to have.**
SFTP is a [supported transfer mechanism](https://restic.readthedocs.io/en/stable/030_preparing_a_new_repo.html#sftp) in Restic, but it is not possible to use it in VolSync: there is no way to configure the SSH key and known_hosts, and the ssh client binary is not present in the Volsync restic container.

**What is the value to the end user? (why is it a priority?)**
Supporting SFTP would allow for more backup destinations, such as rsync.net.

**How will we know we have a good solution? (acceptance criteria)**
- Ability to configure SSH private key and server fingerprint
- Include ssh client included in Restic container

chrishrb:
For those of you, who need this feature: there is a workaround. You can use rclone in restic which supports much more providers, e.g. for sftp you can just use something like this (i know, providing a password is not best practice 😆):

```yaml
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: "${APP}-volsync"
spec:
  secretStoreRef:
    kind: ClusterSecretStore
    name: bitwarden-secretsmanager
  target:
    name: "${APP}-volsync-secret"
    template:
      engineVersion: v2
      data:
        RESTIC_REPOSITORY: "rclone:backup:/k8s-home-backup/${APP}"
        RESTIC_PASSWORD: "{{ .RESTIC_PASSWORD }}"

        # FTPS
        RCLONE_CONFIG_BACKUP_TYPE: sftp
        RCLONE_CONFIG_BACKUP_HOST: "{{ .HOST }}"
        RCLONE_CONFIG_BACKUP_USER: "{{ .USER }}"
        RCLONE_CONFIG_BACKUP_PASS: "{{ .PASSWORD }}"
        RCLONE_CONFIG_BACKUP_SHELL_TYPE: "unix"
        RCLONE_CONFIG_BACKUP_MD5SUM_COMMAND: "none"
        RCLONE_CONFIG_BACKUP_SHA1SUM_COMMAND: "none"
  dataFrom:
    - extract:
        key: 9d80878e-91d8-41e8-8d7e-b2220188317d
---
# yaml-language-server: $schema=https://kubernetes-schemas.pages.dev/volsync.backube/replicationsource_v1alpha1.json
apiVersion: volsync.backube/v1alpha1
kind: ReplicationSource
metadata:
  name: "${APP}"
spec:
[...]

PR #1735, opened 2025-08-14
Add Restic SFTP Support

Jonnobrow:

Adds support for Restic SFTP.

N/A

**Related issues:**

- #671

### DIFF
--- a/bundle/manifests/volsync.clusterserviceversion.yaml
+++ b/bundle/manifests/volsync.clusterserviceversion.yaml
@@ -55,8 +55,8 @@ metadata:
         }
       ]
     capabilities: Basic Install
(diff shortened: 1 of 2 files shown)


tesshuflower:
@Jonnobrow does https://github.com/backube/volsync/pull/1778 help at all with your use-case with restic SFTP?

Jonnobrow:
> @Jonnobrow does https://github.com/backube/volsync/pull/1778 help at all with your use-case with restic SFTP?
> 

Thanks @tesshuflower - I'll have a look into using mover volumes. 

Since these volumes are mounted under `/mnt` I think the restic mover will still need to do some setup for SFTP to work, but it might simplify this solution somewhat.

Issue #1763, opened 2025-09-11
Some options in rclone are hardcoded and cannot be changed


vtj-mizuno:
**Describe the feature you'd like to have.**

Some rclone options (e.g. --transfer, --checksum) are hard-coded in active.sh.
We would like these options to be customizable via environment variables.

**What is the value to the end user? (why is it a priority?)**

These options can be customized to greatly improve rclone performance.
For example, in my environment (tested using rclone docker image), specifying 32 for --transfer reduced the time required for data transfer by more than half.

**Solution plan**

How about reading environment variables as follows?

```
TRANSFERS=${RCLONE_TRANSFERS:-10}
CHECKSUM_FLAG=""
if [[ "${RCLONE_CHECKSUM:-true}" == "true" ]]; then
    CHECKSUM_FLAG="--checksum"
fi

RCLONE_FLAGS_SYNC=(${CHECKSUM_FLAG} --one-file-system --create-empty-src-dirs --stats 20s --transfers ${TRANSFERS})
```

tesshuflower:
I think this is reasonable and something I can implement - `RCLONE_CHECKSUM` and `RCLONE_TRANSFERS` can already be set in the rclone config secret, but currently (as you've seen) they're overridden by the cmd line flag.

So I will most likely depend on the RCLONE_ env vars, and just set defaults to what they are currently, if not specified by the user.

PR #1767, opened 2025-09-17
1763 rclone transfers,  checksum flags

tesshuflower:

RCLONE_ env vars are already passed into the running mover pod, as per documentation here: https://volsync.readthedocs.io/en/stable/usage/rclone/rclone-secret.html#using-rclone-environment-variables-in-rclone-secret

We previously hardcoded `--checksum`, -`-stats`, and `--transfers` flags when running sync and copy commands.

Now update to use RCLONE_ env vars, but set them to our existing defaults if they are not set by the user in the rclone config secret.  This enables users to override those settings if they wish.

Example from testing with logs:

```
2025/09/17 16:15:10 DEBUG : Setting --stats "20s" from environment variable RCLONE_STATS="20s"
2025/09/17 16:15:10 DEBUG : Setting --transfers "10" from environment variable RCLONE_TRANSFERS="10"
2025/09/17 16:15:10 DEBUG : Setting --checksum "true" from environment variable RCLONE_CHECKSUM="1"
```

And after setting RCLONE_TRANSFERS to 19 in the rclone config secret, like this:

```
apiVersion: v1
data:
  RCLONE_TRANSFERS: MTkK
  rclone.conf: <hidden>
kind: Secret
```

running another sync shows this:

```
2025/09/17 17:16:33 DEBUG : Setting --transfers "19" from environment variable RCLONE_TRANSFERS="19"
```

As another example, to disable checksum, set RCLONE_CHECKSUM to 0 or one of f/false/FALSE

**Related issues:**

Fixes: https://github.com/backube/volsync/issues/1763

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Syncthing updated to v1.30.0
 - Rclone updated to v1.69.3
+- Rclone --checksum, --transfers and --stats parameters can now be
+  overridden by RCLONE_ env vars in the rclone config secret
 
 ## 0.13.0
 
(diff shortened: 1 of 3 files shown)


JohnStrunk (approved):
/lgtm

Issue #1424, opened 2024-10-16
Allow volumes and volume mounts to be specified


mimiteto:
**Describe the feature you'd like to have.**
It would be nice to have the option to mount user specified secrets, configmaps (and others) on the replication pods.

**What is the value to the end user? (why is it a priority?)**
As an end user I would be able to use ssh certificates with the `rclone` mover (when using `rclone` with `sftp`).
Currently certificates in `rclone` can be provided only as file paths in their config.
There are possibly other cases where the ability to provide similar configurations is beneficial.

**How will we know we have a good solution? (acceptance criteria)**

**Additional context**

tesshuflower:
We could possibly do something similar to what we did with restic with customCA (that can be specified in a configmap or secret).

Is there a common way of passing this information to rclone, and would this give you what you need? A quick look suggests rclone has a --ca-cert option, but not sure if it works for non HTTPS backends).

The issue with simply mounting things to mover pods is we need to also specify paths so that you can reference them, and also potentially tell rclone where to look for them.  I'd prefer to have defined parameters we pass to rclone if possible.

mimiteto:
About `rclone` - the option there is `--sftp-pubkey-file`.
I proposed a [code change](https://github.com/rclone/rclone/pull/8138) for that specific case, so there is no need to mount additional files (don't know if it will be accepted or not).

About mounting volumes in movers - volumes can be mounted in a specific dir (e.g. `/mnt` or `/data`), allowing users to specify the volumes to be mounted.

mimiteto:
On the topic of `rclone` - the flag I need is `--sftp-pubkey-file`. I proposed a [code change](https://github.com/rclone/rclone/pull/8138) so I don't need to mount anything beyond what is currently available in `volsync`.

On the topic of mounting user specified volumes - I believe this will make `volsync` more flexible.
Currently you are handling each case separately which needs issues like the current one.
Allowing user specified volumes to be mounted will remove the need to open new issues. 
I believe you can just add user specified volumes to a common dir (e.g. `/mnt` or `/mount`).
Let's look at the current replication source
```yaml
apiVersion: volsync.backube/v1alpha1
kind: ReplicationSource
metadata:
  name: user-sync-data-0
  namespace: sync-data
spec:
  rclone:
    copyMethod: Clone
    moverSecurityContext:
      fsGroup: 1000
      runAsGroup: 1000
      runAsUser: 1000
    rcloneConfig: volsync-user-sync-local
[...]

PR #1778, opened 2025-10-02
Addl mover volumes pvc and secret

tesshuflower:
Allows specifying additional mover volumes (secret or PVC type) for replicationsource/replicationdestination.  The associated volume will be mounted to the mover container.

- I didn't bother adding this to rsync, just rsync-tls, restic, rclone and syncthing.
- I did some affinity for PVCs - if we're not already setting nodeSelector/tolerations (for example from being in direct mode), then for each PVC in the moverVolume list, I attempt to get affinity.  Not sure if this is a bit of overkill.

**Related issues:**
https://github.com/backube/volsync/issues/1424

### DIFF
--- a/bundle/manifests/volsync.backube_replicationdestinations.yaml
+++ b/bundle/manifests/volsync.backube_replicationdestinations.yaml
@@ -1391,10 +1391,10 @@ spec:
                       This should only be used by advanced users.
                     items:
                       properties:
-                        name:
+                        mountPath:
                           description: |-
                             Path to give the volume when mounting under /mnt in the mover job pod.
-                            For example if MountPath is 'my-pvc' then this moverVolume will be mounted in the mover pod
+                            For example if mountPath is 'my-pvc' then this moverVolume will be mounted in the mover pod
                             at /mnt/my-pvc
                           type: string
                         volumeSource:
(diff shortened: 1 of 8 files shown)


api/v1alpha1/common_types.go
```diff
@@ -172,4 +172,30 @@ type MoverConfig struct {
...
+	// This should only be used by advanced users.
+	// +optional
+	MoverVolumes []MoverVolume `json:"moverVolumes,omitempty"`
+}
+
+type MoverVolume struct {
+	// Name to give the volume when mounting under /mnt in the mover job pod.
+	// For example if name is 'my-pvc' then this moverVolume will be mounted in the mover pod
+	// at /mnt/my-pvc
+	Name string `json:"name"`
```
JohnStrunk, line 185: What about using a name like `mountPath` similar to the container spec? While this is different than container `mountPath`--- we're forcing it under `/mnt`, I think it's more descriptive than just `name`

tesshuflower:
@JohnStrunk I think this is ready now if you have a chance to review

tesshuflower:
/retest

Both of these appear to be issues with provisioning a cluster for the tests

PR #1809, opened 2025-11-10
add additional tag when pushing to quay

tesshuflower:
To avoid issues with quay cleaning up images after no tags are associated with it anymore.

Can happen for example with latest tag or release-x-y tags

Fixes: https://github.com/backube/volsync/issues/1805

**Related issues:**


tesshuflower:
/close

Closing, this will just end up causing us to save too many images in the end - we would get one for every push to main

PR #1890, opened 2026-01-22
Rclone lostfound

### LINKED ISSUE #1889: RClone - can have issues with lost+found, exclude lost+found with Rclone sync/restore
**Describe the bug**

Unclear exactly when this happens (it doesn't happen in our openshift e2e envs), but perhaps is related to specific kube permissions.  In any case it may be safest to exclude lost+found as it does not look like Rclone does this automatically.

See: https://github.com/backube/volsync/issues/959#issuecomment-3747592823

**Steps to reproduce**

**Expected behavior**

**Actual results**

**Additional context**

tesshuflower:

Excludes lost+found for Rclone

**Related issues:**
Fixes:  https://github.com/backube/volsync/issues/1889

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - moverVolumes updated to allow NFS type volumeMounts
 - Rclone updated to v1.72.1
 
+### Fixed
+
+- Exclude lost+found for Rclone
+
 ## [0.14.0]
 
 ### Changed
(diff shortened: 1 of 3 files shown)


JohnStrunk (approved):
/lgtm

tomaszduda23:
@tesshuflower it is only part of the fix. It fails early with `getfacl`.

VolSync rclone container version: v0.14.0+6cbd2ba
getfacl: Removing leading '/' from absolute path names
getfacl: /data/lost+found: Permission denied

tomaszduda23:
`find "${MOUNT_PATH}" -path "${MOUNT_PATH}/lost+found" -prune -o -print | getfacl - > /tmp/permissions.facl`

Something like this could do.

PR #1943, opened 2026-03-16
Replace perl with python3

falconizmi:

Perl package is bringing unnecessary dependencies and since rhel9, ubi images are pulling rrsync script (which is wrapper script to handle ssh) that is written python3 instead of perl.

And since we dont use perl anywhere else, we are droping perl package, and replacing it with python3 which results in reducing volsync image by nearly half of it size (894 MB -> 512 MB)

**Related issues:**


JohnStrunk:
@tesshuflower @falconizmi Shouldn't this go to main?

tesshuflower:
You're right, this one went directly into the release branch by accident
- John has fixed release-0.16 now to match main (this removes the changes from this PR)
- @falconizmi will make a new PR to get this change into main (which will also go into release-0.16)

PR #1948, opened 2026-03-17
Controller setup & metrics - OpenShift Tls profile compliance

tesshuflower:

controller metrics check for OpenShift central TLS profile - this is an OpenShift initiative to have operators/controllers automatically pickup the central TLS profile spec from the api server and use it.

This PR is step 1  - later will need to make some updates in movers if a TLSProfileSpec is used

- load the tls profile (if OpenShift) and use it for our controller startup (will be used by metrics)
- Use watcher to gracefully restart the operator if the TLS profile is modified
- only call the openshift apis from platform pkg
- add checks to skip TLSProfile settings if not OpenShift
- unit tests
- cache isOpenShift, TLSProfileSpec so multiple calls will not always reload (Openshift wants us to restart the operator anyway if the TLSProfileSpec is changed globally)
- Will also allow movers to call code in platform to get the
        TLSProfileSpec when needed
- Remove old unused get & test for restricted-v2 scc (we have replaced with our own volsync scc)

**Related issues:**

### DIFF
--- a/.ci-scripts/yamlconfig.yaml
+++ b/.ci-scripts/yamlconfig.yaml
@@ -5,8 +5,7 @@ extends: default
 ignore: |
   bundle/**
   config/**
-  internal/controller/test/scc-crd.yml
-  internal/controller/test/populator.storage.k8s.io_volumepopulators.yaml
+  internal/controller/test/**
   custom-scorecard-tests/**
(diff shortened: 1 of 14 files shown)


JohnStrunk (approved):
/lgtm

PR #1955, opened 2026-03-20
OpenShift Tls profile compliance rsync-tls

tesshuflower:

TLS Profile compliance for OpenShift clusters - updates for rsync-tls mover

- destination side only - if the tlsProfileSpec from openshift dictates minTLSVersion and ciphers, these may get used for rsync-tls in setting up the stunnel.conf
Notes:
- stunnel allows for setting TLS 1.3 ciphersuites via the "ciphersuites" setting in stunnel.conf
- Currently it's not possible to set 1.2 ciphers and have it "work" - no PSK ones generally in the lists, and stunnel is always going to default to 1.3 anyway.
- Decision to use TLSv1.3 as the min in OpenShift where TLSPRofileSpec is set, to prevent other (unintended) clients from connecting - we only support our replicationsource as client anyway.
- In the future we can iterate to perhaps allow a different sslversionmin based on profile (and perhaps ciphers), but this would require future TLS versions & support from stunnel/openssl.

**Related issues:**

### DIFF
--- a/internal/controller/mover/rsynctls/mover.go
+++ b/internal/controller/mover/rsynctls/mover.go
@@ -42,6 +42,7 @@ import (
 	volsyncv1alpha1 "github.com/backube/volsync/api/v1alpha1"
 	vserrors "github.com/backube/volsync/internal/controller/errors"
 	"github.com/backube/volsync/internal/controller/mover"
+	"github.com/backube/volsync/internal/controller/platform"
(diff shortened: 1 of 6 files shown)


JohnStrunk (approved):
/lgtm

JohnStrunk (approved):
/lgtm

tesshuflower:
/retest
 
These clusterpool clusters have not been coming up reliably ....

PR #1960, opened 2026-03-24
remove old syncthing option insecureAllowOldTLSVersions

tesshuflower:
- was previously set in our config-template.xml
- Syncthing removed this in v1.29.6 and has forced TLS 1.3 since, so this option now appears to be ignored.

**Related issues:**


JohnStrunk (approved):
/lgtm

PR #1978, opened 2026-04-10
rsync-tls use TLSv1.3 as min, reduce 1.2 ciphers (for stunnel at startup to reduce CPU load)

tesshuflower:
- use ciphers = kPSK also in stunnel.conf to prevent stunnel from loading all PSK ciphers (no 1.2 ciphers will be used anyway, but they still load at startup, and some ciphers will require a DH parameter regen which can cause high CPU load on some systems (can be very slow)

**Related issues:**


JohnStrunk (approved):
/lgtm

PR #1984, opened 2026-04-17
Rsync tls symlink munging

tesshuflower:

Symlinks that are absolute get modified automatically by rsyncd, which results in the target going from something like this;

`mylink -> /opt/linktarget`

to 

`mylink -> opt/linktarget`

This has always been an issue with rsync-tls.  We can use rsync's munging feature to have move the targets to something like `/rsync-munged/opt/linktarget`.  However at the end of the sync on the destination we will need to unmunge to get the proper target back.

This PR updates to find the list of symlinks sent on the source side and sent that list to the dest, which will do the unmunging.

- Advantage is each sync will automatically munge/unmunge the symlinks, so 1st time through will fix ones that were previously incorrect
- Disadvantage is every sync will need to do the unmunge, but we hope normally users should not have huge amounts of symlinks in their pvc.
- By sending the control file from the source, we avoid having to scrape through the entire dest on every sync completion to do unmunging, and only do the specific symlinks.

 - Older k8s were having issues in e2e tests (we have an e2e test in the matrix that runs on 1.20.15 to test our older compatibility).  It seems to be related to the older host-path driver (<1.17.1).  When restoring a PVC, symlinks would be restored with uid/gid of 0:0, which made the compare fail for symlinks.  Workaround is to skip the symlink uid/gid compare only for k8s < 1.24

**Related issues:**

### DIFF
--- a/mover-rsync-tls/client.sh
+++ b/mover-rsync-tls/client.sh
@@ -153,11 +153,22 @@ while [[ $rc -ne 0 && $RETRY -lt $MAX_RETRIES ]]; do
         find "${SOURCE}" -mindepth 1 -maxdepth 1 -printf '/%P\n' > /tmp/filelist.txt
         if [[ -s /tmp/filelist.txt ]]; then
             # 1st run preserves as much as possible, but excludes the root directory
-            rsync -aAhHSxz -r --exclude=lost+found --itemize-changes --info=stats2,misc2 --files-from=/tmp/filelist.txt ${SOURCE}/ rsync://127.0.0.1:$STUNNEL_LISTEN_PORT/data
+            rsync -aAhHSxz -r --exclude=lost+found --itemize-changes --info=stats2,misc2 --files-from=/tmp/filelist.txt ${SOURCE}/ rsync://127.0.0.1:$STUNNEL_LISTEN_PORT/data | tee /tmp/rsync-full.log
         else
             echo "Skipping sync of empty source directory"
         fi
         rc_a=$?
+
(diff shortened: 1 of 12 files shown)


mover-rsync-tls/client.sh
```diff
@@ -153,12 +153,23 @@ while [[ $rc -ne 0 && $RETRY -lt $MAX_RETRIES ]]; do
...
         if [[ -s /tmp/filelist.txt ]]; then
             # 1st run preserves as much as possible, but excludes the root directory
-            rsync -aAhHSxz -r --exclude=lost+found --itemize-changes --info=stats2,misc2 --files-from=/tmp/filelist.txt ${SOURCE}/ rsync://127.0.0.1:$STUNNEL_LISTEN_PORT/data
+            rsync -aAhHSxz -r --exclude=lost+found --itemize-changes --info=stats2,misc2 --files-from=/tmp/filelist.txt ${SOURCE}/ rsync://127.0.0.1:$STUNNEL_LISTEN_PORT/data | tee /tmp/rsync-full.log
         else
             echo "Skipping sync of empty source directory"
         fi
         rc_a=$?
 
+        if [[ $rc_a -eq 0 ]]; then
```
JohnStrunk: Is it intentional to set `rc_a = 0` even if rsync isn't run? The `echo` will set the value of `$?`.

tesshuflower: Good catch, this wasn't intentional and would be an issue if no files to sync ...

JohnStrunk (approved):
/lgtm

PR #2001, opened 2026-05-06
fix release workflow for pkging darwing kubectl bin

tesshuflower:

When running through the release, the darwin version of kubectl compiled, but did not run the correct make target to package as a tar.gz (which the attach to release step is expecting).

**Related issues:**


JohnStrunk (approved):
/lgtm

PR #2004, opened 2026-05-07
Makefile - install native versions of tools

tesshuflower:
- release workflow failed to create darwin version of kubectl as yq was required and GOOS was set previously to darwin, getting error like: "go: cannot install cross-compiled binaries when GOBIN is set"

- Unset GOOS, GOARCH for installing tools like yq

**Related issues:**


JohnStrunk (approved):
/lgtm

PR #2033, opened 2026-06-12
restic optional env vars - only use if in secret

### LINKED ISSUE #2032: Azure Workload Identity no longer working
**Describe the bug**
ReplicationDestination and ReplicationSource no longer auto picks up the correct information for Azure Workload Identity. It seems #1934 broke it.

At least for the ReplicationSource, this worked in 0.15.0. In 0.16.0, manually trying to restore from backup, it is now broken.

**Steps to reproduce**
Set up Azure Workload Identity. Create a service account with federated credentials, and contributor access to a storage account. Make sure the service account is annotated with the manage identity client id.
Set the ReplicationSource/Destination moverServiceAccount to the service account. Use the federated service account. Set moverPodLabels: `azure.workload.identity/use: "true"`

**Expected behavior**
`AZURE_TENANT_ID`, `AZURE_CLIENT_ID` and  `AZURE_FEDERATED_TOKEN_FILE` are all defined as `env:`. No more magic is necessary.

**Actual results**
I tries loading these from a secret.

**Additional context**
When workload identity is enabled, you should not need to do anything else. All necessary secrets are already loaded and defined as the platform intends.

tesshuflower:

Restic mover:

For optional env vars, rather than simply setting in the job/pod, with a mount that is optional, only set them if the var/key exists in the secret. We already have the secret loaded before reconciling the job.

**Related issues:**

Fixes: https://github.com/backube/volsync/issues/2032

Prevents env vars such as `AZURE_TENANT_ID`, `AZURE_CLIENT_ID` and `AZURE_FEDERATED_TOKEN_FILE` from overrwriting env vars that may already be present or injected into the pod when the user doesn't explicitly set them in the restic secret.

### DIFF
--- a/internal/controller/mover/restic/mover.go
+++ b/internal/controller/mover/restic/mover.go
@@ -415,61 +415,10 @@ func (m *Mover) ensureJob(ctx context.Context, cachePVC *corev1.PersistentVolume
 			// location and its password.
 			utils.EnvFromSecret(repo.Name, "RESTIC_REPOSITORY", false),
 			utils.EnvFromSecret(repo.Name, "RESTIC_PASSWORD", false),
-
-			// Optional variables
-			utils.EnvFromSecret(repo.Name, "RESTIC_COMPRESSION", true), // New in v0.14.0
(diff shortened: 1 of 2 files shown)


JohnStrunk (approved):
/lgtm

PR #2087, opened 2026-08-13
require privileged mode for rsync-ssh mover

tesshuflower:

The older rsync mover over ssh always required elevated permissions, and this was allowed to avoid any breakages when movers were introduced that included un-elevated permissions by default.

Now enough time has passed where users should have migrated to the Rsync-TLS mover which does support unprivileged mode.

This change requires the privileged annotation on the namespace to be able to use the rsync ssh mover at all, aligning with the other movers where the annotation is required to run in privileged mode.

**Related issues:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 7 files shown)


tesshuflower:
/cc @JohnStrunk 

Do you mind reviewing?

JohnStrunk (approved):
/lgtm

PR #2090, opened 2026-08-29
Add MoverTolerations to MoverConfig

### LINKED ISSUE #264: RFE: Ability to add taint and tolerations to Backup Jobs to allow them to run on select nodes
It would be nice to be able to provide a node toleration config and node selector labels to control where the jobs run

this would allow us to provision dedicated nodes for backups

codablock:
This PR adds `MoverTolerations` support to `MoverConfig`, which allows to run mover PODs on tainted nodes.

In my case, I have specialised nodes with openebs local-lvm PVCs. I schedule my workloads onto these nodes via nodeSelectors+tolerations. Then I use `ReplicationSource` to periodically backup the PVCs. The movers obviously also have to run on the same nodes, so tolerations on the PODs are mandatory.

**Related issues:**
fixes #264


tesshuflower:
/ok-to-test

Thanks for this @codablock !

PR #2097, opened 2026-09-08
add priorityClassName support for controller on helm chart

### LINKED ISSUE #2095: add priorityClassName passthrough to the helm chart
the helm/volsync chart has no way to set priorityClassName on the operator pod. it exposes nodeSelector, tolerations and affinity already but not this one, so the controller sits at priority 0 no matter what

my use case: volsync runs the backups on my cluster, and on a node squeeze i don't want it to be the first pod evicted

my approach would be a small change: a priorityClassName value defaulting to "", rendered into deployment-controller.yaml next to the existing nodeSelector / affinity / tolerations blocks

i can send the PR if you're ok

JohnMastodon:
adds priorityClassName support in helm chart for the controller pod

**Related issues:**
closes #2095

### DIFF
--- a/helm/volsync/templates/deployment-controller.yaml
+++ b/helm/volsync/templates/deployment-controller.yaml
@@ -99,6 +99,9 @@ spec:
       affinity:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.priorityClassName }}
+      priorityClassName: {{ . | quote }}
+      {{- end }}
       terminationGracePeriodSeconds: 10
       {{- with .Values.tolerations }}
       tolerations:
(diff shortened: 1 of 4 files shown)


helm/volsync/templates/deployment-controller.yaml
```diff
@@ -99,6 +99,9 @@ spec:
       affinity:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.priorityClassName }}
+      priorityClassName: {{ . }}
```
tesshuflower:
```suggestion
      priorityClassName: {{ . | quote }}
```

Could we quote just to ensure we get a string value?

tesshuflower: @JohnMastodon Thanks for this PR - I just have one small comment above - also, you will need to make sure you sign your commit - see: https://github.com/backube/volsync/pull/2097/checks?check_run_id=102285384157

JohnMastodon: @tesshuflower Thanks for the feedback, I included your suggestion in the commit and signed it

